PR #2, opened 2018-11-13
Example output for pubsub

landrito:
# What
This serves as the example output for pubsub.
This includes changes for the following:

- [x] Path Template Handling in Pubsub Clients
- [x] Path Template Input Changes
- [ ] Client changes from #1


examples/google-cloud-pubsub/Gemfile
```diff
@@ -0,0 +1,12 @@
...
+
+gemspec
+
+gem "gcloud-jsondoc",
+    git: "https://github.com/GoogleCloudPlatform/google-cloud-ruby.git",
+    branch: "gcloud-jsondoc"
+
+# WORKAROUND: builds are having problems since the release of 3.0.0
+# pin to the last known good version
+gem "public_suffix", "~> 2.0"
```
blowmage: This is not needed anymore.

examples/google-cloud-pubsub/acceptance/google/cloud/pubsub/v1/publisher_smoke_test.rb
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# AUTO GENERATED BY GAPIC
+
+require "minitest/autorun"
+require "minitest/spec"
```
blowmage: FWIW, "minitest/autorun" requires "minitest/spec", so I don't think this require is necessary.

landrito:
Thanks for the speedy replies @blowmage! Let's keep the conversation about general changes that will be common between #1 and #2 in #1 so they stay in the same place. Once #1 is finalized I move those changes to here. For now let's just focus on the path template stuff for pubsub, then reevaluate the full client after I move #1's changes here.

examples/google-cloud-pubsub/lib/google/cloud/pubsub/v1/publisher_client.rb
```diff
@@ -0,0 +1,713 @@
...
+          ALL_SCOPES = [
+            "https://www.googleapis.com/auth/cloud-platform",
+            "https://www.googleapis.com/auth/pubsub"
+          ].freeze
+
+          # Returns a fully-qualified topic resource name string.
+          # @param project [String]
+          # @param topic [String]
+          # @return [String]
+          def self.topic_path project, topic
```
jbolinger: Are we going to add instance methods for all of these? I suppose we have to keep the class level ones as well.

examples/google-cloud-pubsub/lib/google/cloud/pubsub/v1.rb
```diff
@@ -0,0 +1,221 @@
...
+      # $ gem install google-cloud-pubsub
+      # ```
+      #
+      # ### Preview
+      # #### PublisherClient
+      # ```rb
+      # require "google/cloud/pubsub"
+      #
+      # publisher_client = Google::Cloud::Pubsub::Publisher.new(version: :v1)
+      # formatted_project = Google::Cloud::Pubsub::V1::PublisherClient.project_path(project_id)
```
jbolinger: Are we going to add instance methods for all of these? If so, this should be simplified.

landrito:
# Other Ideas I'm exploring

## Accepting Hashes for path templates
```rb
require "google/cloud/pubsub"

subscriber_client = Google::Cloud::Pubsub::Subscriber.new(version: :v1)
name_ids = {project: "[PROJECT]", snapshot: "[SNAPSHOT]"}
subscription_ids = {project: "[PROJECT]", snapshot: "[subscription]"}
response = subscriber_client.create_snapshot(name_ids, subscription_ids)
```

Problem I am facing with this option is if there is a collision between two paths with their ids.

blowmage:
I think the `#create_subscription` code example could be clearer if the variable names were more descriptive and if we had instance helper methods.

```ruby
require "google/cloud/pubsub"

subscriber_client = Google::Cloud::Pubsub::Subscriber.new(version: :v1)
subscription_path = subscriber_client.subscription_path("[PROJECT]", "[SUBSCRIPTION]")
topic_path = subscriber_client.topic_path("[PROJECT]", "[TOPIC]")
subscription = subscriber_client.create_subscription(subscription_path, topic_path)
```

This code could also be formatted differently to make it even more clear, IMO:

```ruby
require "google/cloud/pubsub"

subscriber_client = Google::Cloud::Pubsub::Subscriber.new(version: :v1)

subscription = subscriber_client.create_subscription(
  subscriber_client.subscription_path("[PROJECT]", "[SUBSCRIPTION]"),
  subscriber_client.topic_path("[PROJECT]", "[TOPIC]")
)
```

Another option to consider is accepting a protobuf resource object, or any object really, that responds to the `#name` method. Then we use the `name` value for the path.

```ruby
require "google/cloud/pubsub"

publisher_client = Google::Cloud::Pubsub::Publisher.new(version: :v1)

topic = publisher_client.get_topic(
  publisher_client.topic_path("[PROJECT]", "[TOPIC]")
)

subscriber_client = Google::Cloud::Pubsub::Subscriber.new(version: :v1)

subscription = subscriber_client.create_subscription(
  subscriber_client.subscription_path("[PROJECT]", "[SUBSCRIPTION]"),
  topic
)
```

jbolinger:
I 👍the formatting and `#name` ideas. 

If we're going to add instance methods for the resources another thing to consider might be a way to aggregate them to keep things well organized, for example how about something like this?

```ruby
subscription_path = subscriber_client.paths.subscription("[PROJECT]", "[SUBSCRIPTION]")
```

quartzmo:
I merged #1 into the `examples` branch, not `master`. Should this PR also be against `examples`? Or did I get it wrong?

blowmage:
Are the changes in this PR needed?

jbolinger:
I think it would be nice to move this into the examples branch for reference. @landrito can you move it at some point?

landrito:
Closing since this is stale.

PR #13, opened 2018-12-19
Update _client.erb

quartzmo:
This addition to `templates/_client.erb` correctly adds the Speech `Recognize` unary method. The docs for the method are incomplete.

This PR is still a work-in-progress, since I need to solve the problem of how to add `require` statements in `clients.erb` that are specified in the child template `_client.erb`. (The solution might be something similar to ActionView's [CaptureHelper](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html).)

Partial visual diff of the speech.rb generated with this PR and the [example speech_client.rb](https://github.com/googleapis/gapic-generator-ruby/blob/examples/examples/google-cloud-speech/output/google-cloud-speech/lib/google/cloud/speech/v1/speech_client.rb):

screen shot 2018-12-19 at 4 29 32 pm

PR #14, opened 2018-12-21
Add actionpack

quartzmo:
This PR converts both `protoc-gen-ruby_gapic_dump` and `protoc-gen-ruby_gapic` to ActionView templates.

[Action View Overview](https://guides.rubyonrails.org/action_view_overview.html)
[Layouts and Rendering in Rails](https://guides.rubyonrails.org/layouts_and_rendering.html)

* Replace TemplateProvider with ActionController::Base.
* Replace Util with ActiveSupport.
* Replace Erubis templates with ActionView.
* Add client template test.


test/templates/client_test.rb
```diff
@@ -0,0 +1,67 @@
...
+      layout: "ruby",
+      assigns: assigns
+    )
+    assert_equal expected, result
+  end
+
+  ##
+  # In the actual generator, the Ruby code in this output is formatted by
+  # Rubocop after it is rendered.
+  def expected
```
jbolinger: Do you think this approach will scale? I think it's fine for now, but I wonder if it'll get awkward as the files get longer and the formatting is odd?

blowmage: I agree this approach fine for now, and also agree that comparing the full output will not scale as the templates change over time.

blowmage:
I like this approach a lot. Being able to use helpers for the utility methods will be much more familiar to most Rubyists.

test/templates/client_test.rb
quartzmo: Yes, this should be considered temporary. But at least the project now has a test! ;)

quartzmo:
I'm working now on a commit fixing `protoc-gen-ruby_gapic_dump`.

test/templates/client_test.rb
quartzmo: What should we call the directory that holds the expected output files for generator tests?

blowmage: Something like `test/expected_output`, I think.

quartzmo:
I added a commit converting `protoc-gen-ruby_gapic_dump` to ActionView templates that I have tested manually. I will add an automated test soon.

quartzmo:
@landrito @jbolinger @blowmage Please see quartzmo/gapic-generator-ruby#2 for an alternative to #13 that utilizes the addition of ActionView templating proposed in this PR.

In particular, it demonstrates the use of [`content_for`](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for) to allow the `client.erb` template to insert `require` statements into the `ruby.erb` layout, above and outside of the module declarations. This is not possible in the current templating in gapic-generator-ruby.

quartzmo:
@landrito In keeping with your inline comment that "users of the generator would be able to add custom templates that generate new files without ever touching any generator code", I will also restore the autoloading and rendering of *all* templates that you had [here](https://github.com/googleapis/gapic-generator-ruby/blob/master/lib/google/gapic/generator/template_provider.rb#L34-L39), the point of which I didn't previously understand.

quartzmo:
@landrito Updated for your previous requests. I will continue to work on the autoloading of templates, but please let me know how it looks so far.

quartzmo:
@landrito @blowmage @jbolinger I have restored @landrito's original design with autoloading and execution of all templates (in a provided path) and file names configured in the templates.

I've tested these changes manually and both `ruby_gapic` and `ruby_gapic_dump` behave as in master branch.

lib/google/gapic/runner/proto_compiler_runner.rb
```diff
@@ -58,25 +60,32 @@ def request
...
-          relative = File.join(*['..']*5, 'templates')
-          File.expand_path(relative, __FILE__)
-        end
-        private :template_path
-
         # Private.
         # The template provider for this run.
-        # @return [Google::Gapic::Generator::TemplateProvider]
+        # @return [ActionController::Base]
         def template_provider
```
landrito: Can we come up with a better name for this?

quartzmo: Changing to `controller` to reflect the actual type.

blowmage: FWIW, I'm fine keeping the current name for now.

quartzmo: @blowmage I already merged, but feel free to change it back.

PR #15, opened 2019-01-03
Update Rubocop conformance

### LINKED ISSUE #9: Add Rubocop
We should add [Rubocop](https://github.com/rubocop-hq/rubocop) to this project and its build to define and enforce a Ruby style guide for this project. A `TargetRubyVersion` and exceptions to default rules should be configured in `.rubocop.yml`. 

Note that #1 added a `.rubocop.yml` file in its example output, but this does not necessarily need to be the same style used for this project, although it might be.

The `.rubocop.yml` file can specify exclusions, such as `test` directory files. Personally I would recommend no exclusions, but we can discuss that decision here in this issue.

blowmage:
Update the code to adhere to the style guide rules.

[fixes #9]

### DIFF
--- a/test/expected_output/client.rb
+++ b/test/expected_output/client.rb
@@ -15,12 +15,12 @@
 # limitations under the License.
 
 module Google
-module Cloud
-module Speech
-module V1
-class SpeechClient
-end
-end
-end
-end
+  module Cloud
+    module Speech
+      module V1
+        class SpeechClient
+        end
+      end
+    end
+  end
 end
--- a/test/expected_output/client.rb
+++ b/test/expected_output/client.rb
@@ -15,12 +15,12 @@
 # limitations under the License.
 
 module Google
-  module Cloud
-    module Speech
-      module V1
-        class SpeechClient
-        end
-      end
-    end
-  end
+module Cloud
+module Speech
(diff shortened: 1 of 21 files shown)


test/expected_output/client.rb
```diff
@@ -15,12 +15,12 @@
...
 
 module Google
-module Cloud
-module Speech
-module V1
-class SpeechClient
-end
-end
-end
-end
```
quartzmo: Can you revert this indentation? The scope of this test does not currently include the `ProtoCompilerRunner#format_file` step, so the correct expectation (currently) should be without indentation. I am getting a `rake test` failure on this branch:

```sh
  1) Failure:
ClientTest#test_client [/Users/quartzmo/code/google/codez/gapic-generator-ruby/test/templates/client_test.rb:35]:
--- expected
+++ actual
@@ -15,13 +15,13 @@
 
 module Google
-  module Cloud
-    module Speech
-      module V1
-        class SpeechClient
-        end
-      end
-    end
-  end
+module Cloud
+module Speech
+module V1
+class SpeechClient
+end
+end
+end
+end
 end
 "
```

quartzmo:
🎉

PR #17, opened 2019-01-04
Add Google Cloud templates

blowmage:
This is an extensive refactor. It builds on the ActionController/ActiveView support and streamlines the Runner and Generator where possible. It allows for templates to be customized, while also allowing for a custom Generator to be used for more advanced usage.

Here are some open issues that will need to be resolved. They can be resolved now or in future PRs.

- [x] Pass the `Google::Protobuf::Compiler::CodeGeneratorRequest` object to the `Google::Gapic::Schema::Api.new` method so it can know about the configuration.
- [ ] Remove environment variables in favor of using `--ruby_gapic_opt` to determine which generator to use and to how to configure it.
- [x] Document what objects are used in each template file using the `<%- assert_locals app, service -%>` helper. (Also, write the helper.)
- [x] Rename `gcloud` to `google_cloud`.
- [ ] Test coverage for the runner and binary.


test/expected_output/gcloud/google-cloud-speech.gemspec
```diff
@@ -0,0 +1,31 @@
...
+# encoding: utf-8
+
+Gem::Specification.new do |gem|
+  gem.name          = "google-cloud-speech"
+  gem.version       = "0.1.0"
+
+  gem.authors       = ["Google LLC"]
+  gem.email         = "googleapis-packages@google.com"
+  gem.description   = "google-cloud-pubsub is the official library for Google Cloud Pub/Sub API."
+  gem.summary       = "API Client library for Google Cloud Pub/Sub API"
```
jbolinger: pubsub?

test/expected_output/gcloud/Rakefile
```diff
@@ -0,0 +1,156 @@
...
+    Rake::Task["smoke_test"].invoke
+  end
+end
+
+# Acceptance tests
+desc "Run the google-cloud-pubsub acceptance tests."
+task :acceptance, :project, :keyfile do |t, args|
+  project = args[:project]
+  project ||=
+    ENV["PUBSUB_TEST_PROJECT"] ||
```
jbolinger: pubsub?

blowmage: Copy/paste. Things are still slowly being un-faked. Getting everything generating and tests passing is a big step though, IMO.

blowmage:
Tests and Rubocop are back to passing.

blowmage:
So, assuming that the Google Cloud generator requires additional configuration, such as gem name and other stuff, would it make sense for the generator to fail if those things are missing? If so, I would really like to identify what additional data can/should be provided, and it might help inform how we break apart the templates.

jbolinger:
I think we should design it such that the generator will produce a working client if the input protos are valid (i.e. with no configuration). Any configuration that is provided either in the proto annotations or as an input option can change the output, but nothing should be required.

For gem name we could either rely on a default or not generate a gem if it's not provided. I'm leaning towards the later.

blowmage:
So have a different set of templates depending on what configuration is passed in? The difference between outputting a full gem and just the client files is pretty big.

jbolinger:
On second thought, I don't know how useful the non-gem version would be so why don't we just start by inferring a gem name if it's not provided. We can always add it later but we will need a gem based output for sure.

templates/default/helpers/api_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module ApiHelper
+  def api_services api
+    api.services.reject { |s| s.name == "Operations" }
```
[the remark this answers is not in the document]
blowmage: Yes. This is definitely faked right now.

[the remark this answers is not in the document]
blowmage: I think we should address this and refactor a more permanent solution in future PRs.

test/expected_output/gcloud/google-cloud-speech.gemspec
blowmage: Fixed.

test/expected_output/gcloud/Rakefile
blowmage: Fixed.

quartzmo:
@jbolinger Do you still want to move more of the content into the `gapic` template set?

blowmage:
Changes would go in the `default` templates, as the `gapic` generator and templates were removed.

quartzmo:
@jbolinger @landrito Any objections to merging this PR?

blowmage:
I think this PR is good now and we should accept and merge it. I think this PR is already so large it is difficult to review, and future additions to it will be more difficult to review than needed. Accepting and merging now will allow for smaller and more frequent changes to improve the templates.

PR #19, opened 2019-01-10
Generate extension gem

blowmage:
Add the ability to generate a new extension gem.


blowmage:
This PR has been rebased on current master to pick up the latest changes and resolve the added conflicts.

jbolinger:
This is a little off-topic, but I don't really like having those `.bin` files be the basis for the tests. Why not use `protoc` and run the tests off of the protos? Or, at least have a script that generates the `bin` files from the protos or a link to the original source?

We can defer this decision but I wanted to bring it up in the context of this split because it might be tricky for someone to do customizations and update their tests if they can't easily see the inputs.

jbolinger:
I'm not suggesting allowing the inputs to change over time. I agree with all of that I just don't like the opaque binary file without the corresponding source (and, to a lesser extent, knowing how the content was generated).

Not saying we need to copy this model but, just to make it concrete, in another generator I did basically the same thing as we're doing here. But, I checked in the binary data (exactly as here) in addition to the source proto and the script used to create it: https://github.com/googleapis/gapic-generator-kotlin/blob/90fe2d45cbed20c199b64b4e14797deb81bc7357/generator/src/test/resources/create-test-data.sh#L17

A permalink to the proto would be enough for me. Just some way to know where this binary thing came from.

blowmage:
Okay, I think a rake task would make the most sense.

jbolinger:
@blowmage are you still iterating on this one or is it ready for review? @landrito this functionality would be great to show off next week.

quartzmo:
I'll review it now.

quartzmo:
When I first looked through this PR, I felt mental stress trying to separate the artifacts for original "generator" from the added artifacts for the "gem generator". I know that files are "generated" by `bin/gapic-generator` but using "generator" for these tasks as well as for the original protoc task is really confusing for me. What do you think about using the term "bootstrap" instead?

gapic-generator/bin/gapic-generator
```diff
@@ -30,10 +53,10 @@ OptionParser.new do |opts|
     options[:export] = export_path
   end
 
-  # opts.on "-g", "--gem name", String,
-  #         "Generate a new rubygem" do |gem_name|
-  #   options[:gem] = gem_name
-  # end
+  opts.on "-g", "--gem name, path", Array, "Generate a new rubygem" do |args|
```
quartzmo: Can the `name` and `path` inputs each have their own option? If not, remove the space, since it doesn't work in practice: `--gem name,path`

gapic-generator/bin/gapic-generator
```diff
@@ -19,6 +19,29 @@ $stderr.sync = true
...
+end
+
+def generate_gem_to gem_name, gem_path
+  if File.exist?(gem_path) && File.file?(gem_path)
+    puts "Cannot generate gem to an existing file path"
+    exit
+  end
+  FileUtils.mkdir_p gem_path
+
+  require "google/gapic/gem_generator"
```
quartzmo: The script currently fails for me at this line:

```sh
hriss-iMac:gapic-generator quartzmo$ ./bin/gapic-generator -g mine,demo
Traceback (most recent call last):
    3: from ./bin/gapic-generator:83:in `<main>'
    2: from ./bin/gapic-generator:38:in `generate_gem_to'
    1: from /Users/quartzmo/.rbenv/versions/2.5.3/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/Users/quartzmo/.rbenv/versions/2.5.3/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- google/gapic/gem_generator (LoadError)
```

quartzmo:
@blowmage This was obviously a lot of work, it's awesome and I'm really impressed!

blowmage:
> When I first looked through this PR, I felt mental stress trying to separate the artifacts for original "generator" from the added artifacts for the "gem generator".

I would never do that intentionally. :trollface: 

I don't like it either, but I couldn't settle on a good name for it. What do you all think about `Google::Gapic::GemBuilder#bootstrap`?

jbolinger:
`bootstrap` seems fitting to me. It does have some overlap with a term we used in the legacy generator, but I don't think we should let that stop us (plus... I can't think of a better alternative).

It's very nice work @blowmage - you should name it 😄

quartzmo:
`GemBuilder#bootstrap` SGTM. `builder` and `bootstrap` (and probably some others) are all fine, just hoping we can avoide the reuse of `generator`.

blowmage:
Updated and rebased. PTAL.

quartzmo:
I'm still getting an error with the gem builder:

```sh
Chriss-iMac:gapic-generator quartzmo$ ./bin/gapic-generator -g mine,demo
Traceback (most recent call last):
	3: from ./bin/gapic-generator:83:in `<main>'
	2: from ./bin/gapic-generator:38:in `build_gem_to'
	1: from /Users/quartzmo/.rbenv/versions/2.5.3/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/Users/quartzmo/.rbenv/versions/2.5.3/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- google/gapic/gem_builder (LoadError)
```

quartzmo:
The rename to `GemBuilder` LGTM.

PR #20, opened 2019-01-11
Cloud template work

blowmage:
This PR is a work in progress to implement much of the faked behavior in the Google Cloud templates. Once we have a working and stable template implementation we can strip out all the Google Cloud specific elements and port the templates to the Default generator.


gapic-generator-cloud/templates/cloud/helpers/method_helper.rb
```diff
@@ -0,0 +1,97 @@
...
+  def method_arg_name arg
+    arg.name
+  end
+
+  def method_arg_type arg
+    if arg.message?
+      message_ruby_type arg.message
+    elsif arg.enum?
+      # TODO: handle when arg message is nil and enum is the type
+      "ENUM(#{arg.enum.name})"
```
blowmage: @landrito I am confused because some Field objects are neither a message or an enum. It would be helpful to go through all these methods with you.

landrito: Ah yeah. The model needs some polishing. We should add some methods in the Field class that check these types: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L137

blowmage: Is this what the Field’s `label` value refers to?

jbolinger: Primitive types aren't messages or enums and they should map to the Ruby equivalent. The `label` field is mostly useful for determining if the field is repeated or singular.

blowmage: If the Field has neither a Message or an Enum, then how do I ask it for the primitive type?

blowmage: Ah, the value is `Field#type`. I'll update this.

blowmage:
FYI, this PR has also been updated on current master.

quartzmo:
@jbolinger I think you had the most recent question on this PR, does it look good to you at this point? Want to approve?

gapic-generator-cloud/Rakefile
```diff
@@ -34,4 +34,34 @@ task :ci do
   Rake::Task[:test].invoke
 end
 
+desc "Generate the expected output for tests"
+task :gen do
+  Rake::Task["gen:templates"].invoke
+end
+namespace :gen do
```
landrito: Discussion for a later PR: Do you think it would be beneficial to expose a rake task as an interface for invoking the generator.

gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
```diff
@@ -54,7 +54,7 @@ def generate
           end
 
           files << cop(gen("credentials.erb",
-                           "lib/#{credentials_file_path @api}.rb",
+                           "lib/#{credentials_file_path @api}",
```
landrito: How are delineating between .rb files and others?

gapic-generator-cloud/templates/cloud/helpers/method_helper.rb
```diff
@@ -0,0 +1,97 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "active_support/inflector"
+
+module MethodHelper
+  def method_name method
```
landrito: hm I'm trying to wrap my head around these helpers. Crazy thought, what if we monkey patched the `Google::Gapic::Schema::{Element}` with these methods rather than prepending each helper with the element name.

gapic-generator-cloud/test/helpers/method/arg_type_test.rb
```diff
@@ -0,0 +1,212 @@
...
+  end
+
+  it "knows the TYPE_DOUBLE type" do
+    type = Google::Protobuf::FieldDescriptorProto::Type::TYPE_DOUBLE # 1
+    arg = field(
+      descriptor(name: "some_method", number: 2, label: 1, type: type),
+      %w[Some Gem some_method]
+    )
+
+    method_arg_type(arg).must_equal String # Decimal
```
jbolinger: Is the `Field#type` not working as expected?

gapic-generator-cloud/Rakefile
jbolinger: 👍 but I don't think it's high priority.

gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
blowmage: That is the difference is designated by using the suffix `_file_path` vs. `_require` (or `_require_path`).

gapic-generator-cloud/templates/cloud/helpers/method_helper.rb
blowmage: The schema structure is not a 1 to 1 for the ruby code structure. I am extremely hesitant to bolt all these ruby code structure methods onto the schema objects. IMO they need to remain separate until we have a firm grasp on what features are needed and how they will be implemented.

Think of the helper methods as well named placeholders for describing the ruby code needs.

gapic-generator-cloud/test/helpers/method/arg_type_test.rb
blowmage: `Field#type` works as expected, I was told to keep `method_arg_type` faked and to not implement it fully until we have a need to finish the implementation. I was hoping the unfinished test would push us towards fully specifying how this should behave, but that will have to wait for a future change. The structure for finishing the implementation is mostly outlined in this test, however.

jbolinger: Oh, that's fine then. I just wanted to make sure we knew how to get the information about the type.

PR #21, opened 2019-01-12
Expose options found in annotations.proto

landrito:
This PR exposes the options found here: https://github.com/googleapis/api-common-protos/blob/input-contract/google/api/annotations.proto

This also vendors the protobuf descriptors and the annotations in order for the annotations to be extracted from the descriptor.


gapic-generator/lib/google/gapic/generators/base_generator.rb
```diff
@@ -15,9 +15,12 @@
 # limitations under the License.
 
 require "action_controller"
-require "protobuf/descriptors"
 require "tempfile"
 
+$LOAD_PATH.unshift ::File.expand_path("../../../../../vendor", __FILE__)
```
blowmage: Can we move this load path hack to only be on schema/api and then load schema/api everywhere we need this?

landrito: I'm a bit unsure of how you wanted it without introducing a load cycle so I isolated the load path to loader.rb.

gapic-generator/lib/google/gapic/schema/loader.rb
```diff
@@ -16,6 +16,10 @@
 
 require "google/gapic/schema/wrappers"
 
+$LOAD_PATH.unshift ::File.expand_path("../../../../../vendor", __FILE__)
```
blowmage: Would `relative_require` be better than a `$LOAD_PATH` change? Why are these files being vendored?

jbolinger: I don't really like having this vendored directory. Do we have any alternatives?

If we must do it I would prefer to have the underlying protos added here as a submodule with a build step to compile them. I know they don't change much, but it will be difficult to know when things get out of date how they were compiled, etc. if we just copy them over.

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -249,6 +261,30 @@ def initialize descriptor, address, docs, input, output
           @output = output
         end
 
+        # return [Array<Google::Api::MethodSignature>] The parameter lists
```
quartzmo: s/return/@return/

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -249,6 +261,30 @@ def initialize descriptor, address, docs, input, output
...
+        #   defined for this method. See `google/api/signature.proto`.
+        def signatures
+          return options[:".google.api.method_signature"] if options
+
+          []
+        end
+
+        # return [Google::Api::OperationData] Additional information regarding
```
quartzmo: s/return/@return/

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -249,6 +261,30 @@ def initialize descriptor, address, docs, input, output
...
+        #   long-running operations.
+        #   In particular, this specifies the types that are returned from
+        #   long-running operations.
+        #   Required for methods that return `google.longrunning.Operation`;
+        #   invalid otherwise.
+        def operation
+          options[:".google.api.operation"] if options
+        end
+
+        # return [Google::Api::HttpRule] The HTTP bindings for this method.
```
quartzmo: s/return/@return/

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -312,6 +348,32 @@ def generate?
           @generate
         end
 
+        # return [Google::Api::Metdata] Packaging information.
+        #   See `google/api/metadata.proto`.
+        def metadata
+          options[:".google.api.metadata"] if options
+        end
+
+        # return [Array<Google::Api::Resource>] A representation of a resource.
```
quartzmo: s/return/@return/

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -312,6 +348,32 @@ def generate?
...
+        #   At a file level, this is generally used to define information for a
+        #   resource from another API, or for a resource that does not have an
+        #   associated proto message.
+        def resources
+          return options[:".google.api.resource_definition"] if options
+
+          []
+        end
+
+        # return [Array<Google::Api::ResourceSet>] A representation of a set of
```
quartzmo: s/return/@return/

gapic-generator/lib/google/gapic/schema/wrappers.rb
landrito: Done.

gapic-generator/lib/google/gapic/schema/wrappers.rb
landrito: Done.

gapic-generator/lib/google/gapic/schema/wrappers.rb
landrito: Done.

gapic-generator/lib/google/gapic/schema/wrappers.rb
landrito: Done

gapic-generator/lib/google/gapic/schema/wrappers.rb
landrito: Done.

gapic-generator/lib/google/gapic/schema/loader.rb
landrito: The only alternative that comes to mind would be to load the protos rather than pre-compiling them but ruby-protobuf doesn't support loading protos at runtime that I know of.

gapic-generator/lib/google/gapic/generators/base_generator.rb
landrito: I simply added the generated files to the lib.

gapic-generator/lib/google/gapic/schema/loader.rb
landrito: I added the protos to the repo and added rake tasks to compile the protos into lib and removed vendor entirely.

WDYT?

jbolinger: Could we `protoc` them as part of a build step and exclude the generated Ruby files from source control?

jbolinger: Ah, I was too slow. Yes, I like that approach better but shouldn't we exclude the generated `.pb.rb` files?

I would also prefer to add the protos as a submodule dependency so it's easier to track them, but we can always do that later.

landrito: That seems reasonable, I'd like some guidance from @quartzmo or @blowmage on the best way to do that.

quartzmo: @jbolinger You mean publishing the generated `.pb.rb` files in another GitHub repo?

quartzmo: Oh nm, I see the protos now. Yes, if those are already published I can see that it might work great to add that repository as a submodule, rather than to copy them.

jbolinger: @quartzmo no, I mean excluding any generated files from source control entirely and adding https://github.com/googleapis/api-common-protos as a git submodule to this project so that there is no duplication.

landrito: I'm gonna give this a shot using `Gem#<post|pre>_install`.

quartzmo: @landrito The build for this repo would use the protos included via git submodule to produce the `.pb.rb` files and include the `.pb.rb` files in the gem for this repo. So what is `Gem#<post|pre>_install` for? Sorry don't follow.

gapic-generator/lib/google/protobuf/descriptor.pb.rb
```diff
@@ -0,0 +1,359 @@
+# encoding: utf-8
```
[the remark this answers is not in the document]
blowmage: Ah, they are in the `protobuf` gem, but not the `google-protobuf` gem. Okay, I get it now.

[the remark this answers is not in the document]
landrito: These are versions of the files exposed by `ruby-protobuf/protobuf`.

Not exactly, If you look into annotations.pb.rb, you can see that the way it is adding the options to is by loading the descriptor classes by requiring the pb files, and then  monkeypatching the descriptor classes into them. If we simply use the descriptor.pb exposed by ruby-protobuf, these won't have those options since the annotations.pb.rb hasn't been loaded.

Furthermore, it would be sensible to think that we should simply be able to load in the annotations.pb.rb file to add those options to the descriptors. This is not the case since the annotations.pb.rb tries to load in the descriptors found at 'google/protobuf/descriptor.pb.rb' which is a different load path than what protobuf is exposing. :/

gapic-generator/lib/google/gapic/schema/loader.rb
quartzmo: So the `.pb.rb` files would not be distributed in the gem for this repo?

blowmage: I would add a rake task to generate the files and include that task in the default build task. Then I would exclude those files by adding `/lib/**/*.pb.rb` to the `.gitignore` file, but include them in the gem by also adding `Dir.glob("lib/**/*.pb.rb")` to the file list in the `gapic-generator.gemspec` file.

landrito: The .pb.rb files will be distributed in the gem but not in the github repo.

Ah that makes sense @blowmage I'll do that.

gapic-generator/lib/google/protobuf/descriptor.pb.rb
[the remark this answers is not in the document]
blowmage: Okay, so this confirms my earlier conclusion: that I don't understand what is going on and I should continue to fake comprehension and trust it will all work out.

_Mike approves the PR_

gapic-generator/lib/google/gapic/generators/base_generator.rb
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 require "action_controller"
-require "protobuf/descriptors"
+require "google/protobuf/compiler/plugin.pb"
```
blowmage: It might be wise to replace the `require "protobuf/descriptors"` statements in the cloud gem as well.

gapic-generator/lib/google/protobuf/descriptor.pb.rb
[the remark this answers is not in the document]
landrito: Yeah this bug was super confusing and took me a whole day to figure out X_X.

gapic-generator/lib/google/gapic/schema/loader.rb
quartzmo: What @blowmage says is exactly what I was thinking too.

jbolinger: I'd prefer to use the submodule approach so that we don't take on a dependency to autosynth. It's a more common practice anyway IMHO.

blowmage: Yep, just an option. One nice thing about committing generated files is that you know what changes happened when. Happy with whatever you settle on.

gapic-generator/Rakefile
```diff
@@ -25,9 +26,41 @@ Rake::TestTask.new :test do |t|
   t.test_files = FileList["test/**/*_test.rb"]
 end
 
+desc "Compile the necessary protobuf files."
+task :compile_protos do
```
landrito: Unfortunately I had to roll my own protoc task since the ruby-protobuf blindly globbed all .proto files and the submoduled repos contain a lot more protos than we need.

gapic-generator/lib/google/gapic/generators/base_generator.rb
landrito: There are none in the cloud gem.

landrito:
I'm very happy with the submoduling solution and not checking in the generated protos. Thanks @jbolinger. I am ready to merge but would like someone else to confirm that:

`bundle install`
`bundle exec rake` 

Works on their machine.

landrito:
Great! Let me update the installation guide in the readme.

gapic-generator/README.md
```diff
@@ -41,6 +41,10 @@ build the generator from scratch.
 ```sh
 $ git clone https://github.com/googleapis/gapic-generator-ruby.git
 $ cd gapic-generator-ruby
+$ git submodule init
+$ git submodule update
+$ bundle install
+$ bundle exec rake
 $ gem build gapic-generator.gemspec
```
quartzmo: Looks like this step is out of date since #18.

```sh
Chriss-iMac:gapic-generator-ruby quartzmo$ gem build gapic-generator.gemspec
ERROR:  Gemspec file not found: gapic-generator.gemspec
```

quartzmo:
Another follow-up request I have for #18 is to restore the top-level `.gitignore` with a line excluding `Gemfile.lock`.

gapic-generator/README.md
landrito: Fixed!

gapic-generator/README.md
```diff
@@ -46,8 +46,9 @@ $ git submodule init
 $ git submodule update
 $ bundle install
 $ bundle exec rake
-$ gem build gapic-generator.gemspec
-$ gem install gapic-generator-0.1.0.gem
+$ cd
```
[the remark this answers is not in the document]
landrito: woops!

landrito:
@quartzmo restored that gitignore.

quartzmo:
More trouble with these steps:

```sh
Chriss-iMac:gapic-generator-ruby quartzmo$ gem build gapic-generator/gapic-generator.gemspec
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    [".rubocop.yml", "CONTRIBUTING.md", "LICENSE", "api-common-protos", "bin/gapic-generator", "bin/protoc-gen-ruby_gapic", "gapic-generator.gemspec", "lib/google/gapic/generator.rb", "lib/google/gapic/generator/version.rb", "lib/google/gapic/generators/base_generator.rb", "lib/google/gapic/generators/default_generator.rb", "lib/google/gapic/runner.rb", "lib/google/gapic/schema.rb", "lib/google/gapic/schema/api.rb", "lib/google/gapic/schema/loader.rb", "lib/google/gapic/schema/wrappers.rb", "protobuf", "templates/default/client.erb", "templates/default/helpers/api_helper.rb", "templates/default/helpers/filepath_helper.rb", "templates/default/helpers/namespace_helper.rb", "templates/default/layouts/_ruby.erb", "templates/default/shared/_client.erb", "templates/default/shared/_license.erb"] are not files
```

landrito:
Thats strange to me. Any idea why that's happening?

quartzmo:
I didn't really look into it, but if I `cd` into the subdir, it seems to work fine, so should we go with that for each gem?

```sh
Chriss-iMac:gapic-generator-ruby quartzmo$ cd gapic-generator
Chriss-iMac:gapic-generator quartzmo$ gem build gapic-generator.gemspec 
  Successfully built RubyGem
  Name: gapic-generator
  Version: 0.1.0
  File: gapic-generator-0.1.0.gem
```

quartzmo:
Also, could you please remove the `.gitignore` I asked you to restore? It didn't have `Gemfile.lock` in it, and I think we can discuss this issue with blowmage outside of this PR. Sorry!

PR #29, opened 2019-01-30
Dynamically Generate Modules for client and credentials

### LINKED ISSUE #26: Dynamically Generate Modules
## What
Inside [client.erb#L18](https://github.com/googleapis/gapic-generator-ruby/blob/master/gapic-generator-cloud/templates/cloud/client.erb#L18), the modules that the client is defined in are hard coded. These modules should be determined by the namespace of the service.

Similarly in [credentials.erb#L9](https://github.com/googleapis/gapic-generator-ruby/blob/master/gapic-generator-cloud/templates/cloud/credentials.erb#L9)

## Data
The data for the namespace can be access through the `Google::Gapic::Schema::Service#address` attribute. This address will return a list of strings in which the last string is the name of the service.

For speech this would look like: 
```rb
service.address # => ["google", "cloud", "speech", "v1", "Speech"]
```

## How
A solution would be to use [defaults/layouts/_ruby.erb](https://github.com/googleapis/gapic-generator-ruby/blob/master/gapic-generator/templates/default/layouts/_ruby.erb), passing in the namespaces and yielding the client body.

landrito:
Resolves #26


gapic-generator/templates/default/layouts/_ruby.erb
```diff
@@ -1,13 +1,18 @@
...
-
-<% namespaces.each do |namespace| %>
-module <%= namespace.upcase_first %>
-<% end %>
-<%= yield %>
-<% namespaces.count.times do %>
-end
 <% end %>
+<%- namespaces.each_with_index do |n, i| -%>
+<%= indent "module #{n.upcase_first}", i*2 %>
```
jbolinger: Will this work? 

It needs to be able to handle overrides via the `ruby_package` (example: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto)

Feel free to open an issue and defer it for this PR - this caused a lot of pain the current generator so I don't want to let it get out of hand again.

gapic-generator-cloud/templates/cloud/client.erb
```diff
@@ -1,121 +1,8 @@
...
-            end
-          end
-        end
-      end
-    end
-  end
-end
-
-<%= render partial: "client/helpers", locals: { service: service} %>
+<%= render partial: "client/client",
```
quartzmo: What is the point of delegating the entire output of this template to a partial?

quartzmo: Is it just to use the `ruby` layout? Is there a better way?

gapic-generator-cloud/expected_output/speech/lib/google/cloud/speech/v1/speech_client.rb
```diff
@@ -43,7 +40,7 @@ class SpeechClient
           DEFAULT_SERVICE_PORT = 443
 
           # The default set of gRPC interceptors.
-          GRPC_INTERCEPTORS = []
+          GRPC_INTERCEPTORS = [].freeze
```
blowmage: I don't understand why this is being frozen. The whole point of this constant is to be a place to define the interceptors for the clients.

quartzmo:
`rake ci` is failing with:

```sh
33 files inspected, 30 offenses detected
RuboCop failed!
```
It appears that only 1 of these rubocop errors is from this branch, but we need to do some cleanup at some point.

gapic-generator-cloud/templates/cloud/client.erb
blowmage: It looks like it is moved to a partial to use the layout. I think being more granular with partials would be okay, as it would help maintaining a consistent indentation.

gapic-generator-cloud/templates/cloud/client/_client.erb
```diff
@@ -0,0 +1,111 @@
...
+        Google::Gax::Grpc::STATUS_CODE_NAMES,
+        timeout,
+        errors: Google::Gax::Grpc::API_ERRORS,
+        metadata: headers
+      )
+    end
+  end
+end
+
+<% @footer = capture do %>
```
blowmage: why are you capturing here?

gapic-generator/templates/default/layouts/_ruby.erb
landrito: Yep, I can have the api schema determine the address from the ruby_package option. Just haven't gotten around to it.

gapic-generator-cloud/templates/cloud/client.erb
landrito: Yep it is simply to use the layout. I do agree that it would be nice to find a better way to do this though. I'll think about it.

landrito:
@quartzmo running `bundle exec rake ci` yielded me different rubocop failures. I fixed the things. First commit is rubocop failures related to this PR and the second for already present failures.

gapic-generator-cloud/expected_output/speech/lib/google/cloud/speech/v1/speech_client.rb
landrito: Fixed.

quartzmo:
@landrito I ran `rake ci` at the top-level, which first runs `ci` for `gapic-generator` (result: `33 files inspected, 29 offenses detected`). Thanks for fixing rubocop for `gapic-generator-cloud`.  I will clean up `gapic-generator` errors in a separate PR right now.

quartzmo:
Ooops, the rubocop errors I see locally in `gapic-generator` are all in the output of the `GemBuilder`, not in git-managed files. Sorry, false alarm.

PR #32, opened 2019-02-01
Split the method definition into multiple templates

blowmage:
This is a sister PR to #33 showing two different approaches to managing the complexity in generating a GAPIC client method. This PR identifies 5 different types of methods and uses 5 different method definition templates. The logic of what to include in each type of method is localized to the individual method templates. There is duplication across the templates.


blowmage:
Closing in favor of #33.

PR #34, opened 2019-02-19
Update client_test.erb

quartzmo:
Dynamically generates tests for service methods. Currently supports the unary, LRO and bidi streaming methods in the Speech client.


gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -170,48 +78,85 @@ describe Google::Cloud::Speech::V1::SpeechClient do
...
       )
+<% end %>
 
       # Mock Grpc layer
+<% if method_lro? method %>
       mock_method = proc do |request|
-        assert_instance_of Google::Cloud::Speech::V1::LongRunningRecognizeRequest, request
-        assert_equal Google::Gax.to_proto(config, Google::Cloud::Speech::V1::RecognitionConfig), request.config
-        assert_equal Google::Gax.to_proto(audio, Google::Cloud::Speech::V1::RecognitionAudio), request.audio
+        assert_instance_of <%= ruby_namespace namespaces %>::<%= method_name.camelize %>Request, request
```
jbolinger: We should get this from the proto (can't assume that it follows this naming pattern).

quartzmo:
> There are a lot of things in this template that won't generalize (totally fine for this PR).
> One in particular that it might be worth tackling now though is setting up a proper way to map protobuf to Ruby type names

@jbolinger Ah of course. I'll try to follow that approach everywhere I can. Please point out anything else that you see here that won't generalize if you get the chance.

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -170,48 +78,85 @@ describe Google::Cloud::Speech::V1::SpeechClient do
...
+<% if method_lro? method %>
           assert_equal expected_response, response.response
+<% elsif method_streaming_bidi? method %>
+          assert_equal 1, response.count
+          assert_equal expected_response, response.first
+<% else %>
+          assert_equal expected_response, response
+
+          # Call method with block
+          client.recognize config, audio do |resp, operation|
```
jbolinger: method arguments here won't generalize

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -220,47 +165,52 @@ describe Google::Cloud::Speech::V1::SpeechClient do
...
         done: true,
         error: operation_error
       )
 
       # Mock Grpc layer
       mock_method = proc do |request|
-        assert_instance_of Google::Cloud::Speech::V1::LongRunningRecognizeRequest, request
-        assert_equal Google::Gax.to_proto(config, Google::Cloud::Speech::V1::RecognitionConfig), request.config
-        assert_equal Google::Gax.to_proto(audio, Google::Cloud::Speech::V1::RecognitionAudio), request.audio
+        assert_instance_of <%= ruby_namespace namespaces %>::<%= method_name.camelize %>Request, request
```
jbolinger: wont' generalize

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -269,91 +219,35 @@ describe Google::Cloud::Speech::V1::SpeechClient do
...
-
-      # Mock auth layer
-      mock_credentials = MockSpeechCredentialsV1.new "streaming_recognize"
-
-      Google::Cloud::Speech::V1::Speech::Stub.stub :new, mock_stub do
-        Google::Cloud::Speech::V1::Credentials.stub :default, mock_credentials do
-          client = Google::Cloud::Speech.new version: :v1
+      <%= ruby_namespace service.address %>::Stub.stub :new, mock_stub do
+        <%= ruby_namespace namespaces %>::Credentials.stub :default, mock_credentials do
+          client = <%= ruby_namespace service.address[0..-3] %>.new version: :<%= namespaces.last.downcase %>
```
jbolinger: won't generalize

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -269,91 +219,35 @@ describe Google::Cloud::Speech::V1::SpeechClient do
...
 
           # Call method
           err = assert_raises Google::Gax::GaxError do
-            client.streaming_recognize [request]
+<% if method_streaming_bidi? method %>
+            client.<%= method_name %> [request]
+<% else %>
+            client.<%= method_name %> config, audio
```
jbolinger: won't generalize

quartzmo:
@jbolinger Great, thank you for marking these!

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -1,22 +1,21 @@
...
-
 require "minitest/autorun"
 require "minitest/spec"
 
 require "google/gax"
 
-require "google/cloud/speech"
-require "google/cloud/speech/v1/speech_client"
-require "google/cloud/speech/v1/cloud_speech_services_pb"
+require "<%= ruby_path service.address[0..-3] %>"
```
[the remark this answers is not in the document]
blowmage: IMO this should be `require <%= gem_require %>`.

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -1,22 +1,21 @@
...
+require "<%= ruby_path service.address %>_client"
+require "<%= ruby_path namespaces %>/cloud_<%= service.name.underscore %>_services_pb"
```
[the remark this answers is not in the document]
blowmage: IMO this should be `require <%= client_require(service) %>`.

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -1,22 +1,21 @@
...
 require "google/longrunning/operations_pb"
 require "google/longrunning/operations_services_pb"
 
-class CustomTestErrorV1 < StandardError; end
+class CustomTestError<%= namespaces.last.upcase_first %> < StandardError; end
```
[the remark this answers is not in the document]
blowmage: `<%= custom_error(service) %>`

gapic-generator-cloud/templates/cloud/client_test.erb
quartzmo: I can't think of how else to generalize `client.recognize` except with `client.<%= method_name %>`. Can you explain a bit?

jbolinger: That part is okay. It's the method arguments (`config, audio`) that will have to change.

gapic-generator-cloud/templates/cloud/client_test.erb
```diff
@@ -170,48 +78,85 @@ describe Google::Cloud::Speech::V1::SpeechClient do
       }
       uri = "gs://bucket_name/file_name.flac"
       audio = { uri: uri }
+<% end %>
 
       # Create expected grpc response
       expected_response = {}
       expected_response = Google::Gax.to_proto expected_response,
-                                               Google::Cloud::Speech::V1::LongRunningRecognizeResponse
+                                               <%= ruby_namespace namespaces %>::<%= method_name.camelize %>Response
```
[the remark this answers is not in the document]
quartzmo: I'm running into difficulty with this one, because the method proto does not describe the nested response (`Google::Cloud::Speech::V1::LongRunningRecognizeResponse`), only the `output_type=".google.longrunning.Operation" `.

Well, this isn't entirely true: The proto contains:

```
leading_comments=" Performs asynchronous speech recognition: receive results via the\n google.longrunning.Operations interface. Returns either an\n `Operation.error` or an `Operation.response` which contains\n a `LongRunningRecognizeResponse` message.\n"
```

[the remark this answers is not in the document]
jbolinger: This is covered by the new config spec (you can check the doc linked from our meeting invite or ping me and I'll send you a link).

However, to cause more confusion, the implementation of those configuration options has changed a bit in the past couple of weeks and the docs have not been kept up to date.

On the plus side, it's not too complicated. Here's a proto that I use for tests in another generator that I think will help you understand all the available options (this covers almost everything with the exception of the oauth scopes and resource names):
https://github.com/googleapis/gapic-generator-kotlin/blob/master/generator/src/test/resources/proto/google/example/test_annotations.proto

I think it'll really help if we can get the test infrastructure @blowmage is working on out. Then we can double check that we have all these configuration options covered in our "garbage" protos and that we all understand them (also there are some gaps - fortunately this case isn't one of them).

So, to generate this particular method, we would expect the proto to have this annotation on it:
https://github.com/googleapis/gapic-generator-kotlin/blob/72b36a26fd1dc2f03aa4c82697203bd86d66bf66/generator/src/test/resources/proto/google/example/test_annotations.proto#L51

If it has that annotation you can extract the type info that you need to generate this method and document the return type correctly. If it does not have that annotation then the method should be treated as a regular method and simply return a `Google::Longrunning.Operation` with no special treatment.

Finally, please keep in mind that that this whole configuration thing is still in flux. If you check our `.gitmodules` we have a dependency on the protos that defines the configuration at `gapic-generator/api-common-protos`, which is now out of date.

quartzmo:
@blowmage I moved argument values to `MethodPresenter`. I think this PR is probably now at a good place to merge, although I'm happy to continue making updates if you see anything else you think would be good to do at this time.

quartzmo:
Top-level `rake ci` is passing for this PR.

PR #35, opened 2019-02-22
Add placeholder for showcase functional test suite

jbolinger:
This adds a new project for the showcase conformance tests that will be added for each generator.

Note that this is just the scaffolding. The placeholder test currently fails and invoking the generator is currently failing with:
```
/Users/jbolinger/Development/gapic-generator-ruby/gapic-generator/lib/google/gapic/runner.rb:37:in `run': uninitialized constant Google::Gapic::Schema (NameError)
	from /Users/jbolinger/Development/gapic-generator-ruby/gapic-generator/bin/protoc-gen-ruby_gapic:24:in `<top (required)>'
	from /Users/jbolinger/.rvm/gems/ruby-2.4.1/bin/protoc-gen-ruby_gapic:23:in `load'
	from /Users/jbolinger/.rvm/gems/ruby-2.4.1/bin/protoc-gen-ruby_gapic:23:in `<main>'
--ruby_gapic_out: protoc-gen-ruby_gapic: Plugin failed with status code 1.
```

I will update these tests as soon as the generator is capable of generating a functional client.

Also, let me know what you think of using docker locally for the tests. I thought this would be a handy way to do it but you may have better ideas. I'll also need to add something so it doesn't try to start the server this way in CI once we have that setup.


gapic-showcase-test/showcase-protos/google/api/client.proto
```diff
@@ -0,0 +1,166 @@
...
+//
+//     option (google.api.package) = {
+//       title: "Pub/Sub"
+//       namespace: ["Google", "Cloud"]
+//       version: "v1"
+//     };
+//
+// If this is set, then the `title`, `namespace`, and `version` fields
+// should all be set.
+message Package {
```
blowmage: Is this the same file that is available at `shared/api-common-protos/google/api/annotations.proto`?

jbolinger: Yes, but that stuff was moved. (i.e. most of the things in `annotations.proto` were moved to `client.proto` a few weeks ago).

blowmage:
This is good to merge. PTAL.

jbolinger:
Thanks for updating this! 

We only need the `echo` client for now, but putting them all in here as you've done is fine so long as you're okay with a the small amount of extra churn it might create by putting them all in place now.

(I can't approve since I started it though)

quartzmo:
@jbolinger I approved this yesterday, do you want to merge?

jbolinger:
Oh, sorry about that @quartzmo - thanks for the reminder.

PR #37, opened 2019-03-01
Restructure project layout

blowmage:
This PR is all (most) of the changes needed before adding the garbage protobuf files. I decided to extract this PR from the upcoming garbage proto/annotations PR to make the changes easier to review. Apologies for this being so large.

This PR provides an automated way to create binary input files. And it is a pretty large restructure of the project to ensure that binary input files referencing configuration files paths to configuration files can be used in the gems and still find the configuration files. It does other things like moving the api-common-protos and protobuf submodules and replacing the duplicate input files with common files using symlinks.

The speech binary file has been updated with the latest protos, and the vision binary file has also been added.

The executable binary files have been refactored to read from `STDIN` and write to `STDOUT` instead of the runner. The binary files now create the `CodeGeneratorRequest` object and pass it to the runner. The customized generator gems like `cloud` now pass the preferred generator type to the runner.

Environment variables have been removed and the generator type and template path is now configured using the protoc options arguments.


common/README.md
```diff
@@ -0,0 +1,79 @@
...
+$ cd /usr/src/protoc/
+$ unzip protoc-${PROTOBUF_VERSION}.zip
+$ rm protoc-${PROTOBUF_VERSION}.zip
+
+# Link the protoc to the path.
+$ ln -s /usr/src/protoc/bin/protoc /usr/local/bin/protoc
+$ mkdir -p /protos/
+
+# Move the common protobuf files to the local include folder.
+$ cp -R /usr/src/protoc/include/* /usr/local/include/
```
jbolinger: I know this isn't new, but since we're reorganizing, could we add the `grpc-tools`gem as a development dependency or something similar that would make this setup easier?

common/Rakefile
```diff
@@ -0,0 +1,65 @@
...
+    # TODO: set output to /dev/null?
+    "--ruby_gapic_out=./output/protos/#{name}",
+    # TODO: invoke a custom runner?
+    # "--ruby_gapic_opt=generator=cloud",
+    # TODO: invoke a custom template path?
+    # "--ruby_gapic_opt=templates=../gapic-generator-cloud/templates",
+    "--ruby_gapic_opt=binary_output=./input/#{name}.bin"
+  ].join " "
+
+  `#{protoc_cmd}`
```
jbolinger: I'm curious why you prefer this over `sh` or something more explicit?

common/README.md
blowmage: This is a copy of the gapic-generator README. I don't know anything about this. @landrito would know more than me.

I can revise the README to be more specific to how the directory is being used.

common/Rakefile
blowmage: Mostly habit. The backtick will return the output from the command, so if you wanted to have the task print the output you can pass the output to `puts`:

```ruby
puts `#{protoc_cmd}`
```

Using backtick everywhere also makes it easy to search for instances of where a command is run. You don't have to search for `sh` and `system` and `exec` and `%x` and `Open3.popen3` and ...

common/README.md
jbolinger: Let's deal with it later then. I don't really like asking people to do things in `/usr/src/`, etc. but it's a documentation thing we can address at some point.

blowmage:
> How about keeping the name `expected_output` in both places?

I named them differently in `common` to see how it felt. I think I prefer the different names in `common` than in the gems because the way the directories are used is different. I found I liked having the `output` in the `common` directory as it makes it easier to scroll through all the gem's expected output using my editor in one place. Its not technically necessary, but was one of the thing I tried and liked.

I don't think all of the names we've chosen are perfect. I still feel like there is a lot of room for naming improvement, and nothing is stopping us from changing the names. For example, I'm not entirely happy with Generator vs. GemBuilder. That said, I usually find that the longer we live with not-quite-right names the easier it is to come to an agreement on what the proper name is. So I'm happy to let this sit for now and trust the process.

We could open an issue to revisit these names before this is shipped.

common/README.md
blowmage: TBH I didn't like it either, and had the same concerns. But this file here is a copy/paste. I'll update the README and take this out.

jbolinger:
Sure, you can make the call on the name thing. I don't think it's a big deal it just wasn't what I had expected.

blowmage:
I think this is just my process too. Poor @quartzmo has suffered through lots of this type of churn as I try to identify a design. I don't feel like I have a solid handle on what the final implementation will look like yet, so I end up throwing stuff at the wall and see what sticks.

PR #38, opened 2019-03-06
Refactor some names

blowmage:
Rename common to shared
Rename Schema::Api#parameters to Schema::Api#protoc_options
Add Schema::Api#configuration
Add configuration option and config files
Add console tasks, with schema helper method
Add update task, to bundle update all gems and shared


gapic-generator-cloud/Rakefile
```diff
@@ -46,6 +46,25 @@ namespace :gen do
...
+task :console do
+  require "irb"
+  require "irb/completion"
+  require "pp"
+
+  $LOAD_PATH.unshift "lib"
+
+  require "google/gapic/schema"
+  def schema service
+    proto_input = File.binread "proto_input/#{service}.bin"
```
jbolinger: This looks like a nice way to explore the schema object. Could we take a proto as input here and compile it or do you think that's overkill? Is there another reason to want to use a `.bin` file?

blowmage: Right now there is still some configuration specific to each service used to build the binary file in the shared Rakefile. I suppose we could put that somewhere where it can be easily used across all the directories. But I think it is easier to load a file and inspect it.

jbolinger: That makes sense for now. I could see something like this being pretty helpful to help get someone started, but let's not complicate it now while the config is still in motion.

PR #41, opened 2019-03-07
Add garbage protos

blowmage:
These protos are a way to cover all edge cases not covered in the other protos. They can be used to test all field types, all RPC method types, all annotation, etc.


blowmage:
This PR doesn't include 1) the changes necessary to get the updated annotations working properly 2) updates to the cloud generator to get it to build the garbage protos 3) updated helper tests using the additional fields these protos provide. But the protos do include the updated annotations and are compiling without warning. I expect the garbage protos to be added to over time as new areas for test coverage are identified.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+import "google/api/annotations.proto";
+import "google/api/client.proto";
+import "google/api/resource.proto";
+import "google/longrunning/operations.proto";
+import "google/protobuf/timestamp.proto";
+
+option (google.api.client_package) = {
+  title: "Testing Garbage"
+  namespace: ["Google", "Cloud"]
+  version: "v1"
```
jbolinger: Same. I'd prefer to use something strange that does not look like a version or follow a typical version regex.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+};
+
+// Endless trash
+service GarbageService {
+  // The address where this API is typically served.
+  option (google.api.default_host) = "garbage.googleapis.com";
```
jbolinger: Same. More random trash and less Google references.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+
+  // The OAuth scopes needed to make requests to this API.
+  option (google.api.oauth_scopes) = "https://cloud.google.com/auth/cloud-platform,"
```
jbolinger: Same. Need more trash here.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+    };
+    // Once the operation is complete, the ultimate result
+    // will be a LongRunningGarbageResponse.
+    option (google.longrunning.operation_info) = {
+      response_type: "google.garbage.v1.LongRunningGarbageResponse"
+      metadata_type: "google.garbage.v1.LongRunningGarbageMetadata"
+    };
+  }
+
+  // This method split the given content into words and will pass each word back
```
jbolinger: Why do these methods have the same comments as showcase (copy/paste)? Some comments are fine but we probably shouldn't use the same one here.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
+syntax = "proto3";
+
+package google.garbage.v1;
```
jbolinger: Let's use a namespace that doesn't include `google` and doesn't end in a version.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+message SimpleGarbageItem {
+  string garbage = 1 [
+    (google.api.resource_reference) = "SimpleGarbage"];
+  oneof item {
+    string a = 2;
+    int32  b = 3;
+  }
+}
+
+message TypicalGarbage {
```
jbolinger: Maybe add a nested type or two for completeness?

quartzmo:
Now that I see this, I'm wondering what makes it different from gapic-showcase `EchoClient`?

jbolinger:
@quartzmo it's probably not so different as we get started but it will be once we really get going (and testing). Showcase is sort of a normal API and we'll want a proto that we can put all sorts of strange things in to test edge cases, etc. that we can't put in showcase. Hence, this garbage :)

shared/output/gapic/templates/garbage/google/garbage/v1/garbage_service.rb
```diff
@@ -0,0 +1,352 @@
...
+require "pathname"
+
+require "google/gax"
+require "google/gax/operation"
+require "google/longrunning/operations_client"
+
+require "google/cloud/garbage_service/v1/cloud_garbage_service_pb"
+require "google/cloud/garbage_service/v1/credentials"
+
+module Google
```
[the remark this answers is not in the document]
blowmage: One is the file to compare for tests, one is the output from generating the binary file. I have a comment in the binary file rake task to send the output to `/dev/null` instead.

gapic-generator-cloud/Rakefile
```diff
@@ -43,6 +43,7 @@ namespace :gen do
   task :templates do
     generate_cloud_templates "speech"
     # generate_cloud_templates "vision"
+    # generate_default_templates "garbage"
```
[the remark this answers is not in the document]
blowmage: The cloud generator currently breaks when trying to generate anything other than speech, because so much of it is faked out and specific to speech. The gapic generator is able to build all the inputs, so it isn't commented out.

shared/protos/garbage/garbage.proto
blowmage: How do you mean? Isn't `GarbageMap` a nested type?

blowmage:
@quartzmo There are many things that this has that showcase does not, such as using all available field types, defining `oneof` fields, using `resource`, `resource_reference`, and `resource_set` annotations, etc...

shared/protos/garbage/garbage.proto
jbolinger: No, I mean something like:

```proto
message Foo {
    message Bar {
        // ....
    }
    // ....
}
```

It's not something we need to do right now, but at some point we will want to make sure the generator can handle nested types correctly.

shared/protos/garbage/garbage.proto
```diff
@@ -0,0 +1,234 @@
...
+
+import "google/api/annotations.proto";
+import "google/api/client.proto";
+import "google/api/resource.proto";
+import "google/longrunning/operations.proto";
+import "google/protobuf/timestamp.proto";
+
+option (google.api.client_package) = {
+  title: "Testing Garbage"
+  namespace: ["Google", "Cloud"]
```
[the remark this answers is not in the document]
blowmage: Trashed it up.

shared/protos/garbage/garbage.proto
blowmage: Trashed it up.

shared/protos/garbage/garbage.proto
blowmage: Trashed it up.

shared/protos/garbage/garbage.proto
blowmage: Trashed it up.

shared/protos/garbage/garbage.proto
blowmage: Removed.

shared/protos/garbage/garbage.proto
blowmage: Trashed it up.

shared/output/gapic/templates/garbage/google/garbage/v1/garbage_service.rb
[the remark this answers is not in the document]
blowmage: These files are now removed.

shared/protos/garbage/garbage.proto
blowmage: Oooohhh, I didn't even realize we could do that...

blowmage: Added.

blowmage:
Endless trash!

blowmage:
I didn't add the `google.api.method_signature` annotations because we already discussed changing what is accepted to the client methods. It would be nice to allow users to provide either the request object as a positional argument, or the request object values as named arguments. Because of this, we won't be generating method signatures using these annotations. But, we could generate code examples using them.

What do you all think? Is it worth adding `google.api.method_signature` annotations to the garbage proto?

jbolinger:
I would vote for adding a `method_signature` somewhere even if we completely ignore it in the generator. 

It'll serve as explicit documentation that this annotation has no effect and since it'll be commonly found in the input it wouldn't be bad to have some simple test that shows it doesn't do anything. Not a high priority item by any means though.

blowmage:
K, added some `method_signature` and `field_behavior` annotations.

jbolinger:
@blowmage the extra content looks good. Are you planning to do anything more here?

blowmage:
I don't have anything else for this PR.

PR #42, opened 2019-03-12
Add CircleCI config

jbolinger:

PR #43, opened 2019-03-12
Fix CircleCI build and add badge

jbolinger:
The `config.yml` add in #42 doesn't seem to install all dependencies required by the gems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # API Client Generator for Ruby
 
+[![CircleCI](https://circleci.com/gh/googleapis/gapic-generator-ruby.svg?style=svg)](https://circleci.com/gh/googleapis/gapic-generator-ruby)
+
 Create Ruby clients from a protocol buffer description of an API.
 
 **Note** This project is a preview. Please try it out and let us know what you think,


jbolinger:
@blowmage do you know why this doesn't install everything (in this case it's `minitest-focus`)?

blowmage:
The config is using both `vendor/bundle` and `.vendor/bundle`, is that perhaps part of the problem?

.circleci/config.yml
```diff
@@ -15,17 +15,15 @@ jobs:
...
       - restore_cache:
           keys:
-            - v1-dependencies-{{ checksum "Gemfile.lock" }}
-            # fallback to using the latest cache if no exact match is found
-            - v1-dependencies-
+            - v1-dependencies-generator
       - run:
           name: install dependencies
           command: |
             bundle install --jobs=4 --retry=3 --path vendor/bundle
```
blowmage:
```suggestion
            bundle install --jobs=4 --retry=3 --path .vendor/bundle && bundle clean
```

blowmage: Added `bundle clean` because of the note [here](https://circleci.com/docs/2.0/caching/#bundler-ruby).

blowmage: Also fixes the path so bundler installs dependencies in the same directory that is being cached.

jbolinger: Interesting. I wonder why they don't suggest that during the setup. I just copy pasted this config. I suspect that api common protos repo I setup awhile back has this same cache path issue.

jbolinger: This still doesn't install `minitest-focus` when I run it locally. If I `cd` into `gapic generator` and do an install it's fine, but it doesn't work when doing an install at the root only.

Issue #44, opened 2019-03-13
Allow running the plugin from github


jbolinger:
I was trying to find a simple way to let people try out the generator and write a test script without downloading anything. This is the best method that I have found so far:

`Gemfile`
```ruby
#gem "gapic-generator", github: "googleapis/gapic-generator-ruby", branch: "master"
gem "gapic-generator", path: "../gapic-generator-ruby/gapic-generator"

gem "bundler", "~> 1.15"
gem "rake", "~> 10.0"
gem "grpc-tools", "~> 1.18"
```

`Rakefile`
```ruby
desc 'Generate showcase.'
task :generate_showcase do
  FileUtils.mkdir 'lib'

  command = []
  command << 'grpc_tools_ruby_protoc'
  command << '--ruby_out=lib'
  command << '--ruby_gapic_out=lib'
  command << '--proto_path ./proto/api-common-protos'
  command << '--proto_path ./proto/showcase'
  command << './proto/showcase/v1alpha3/echo.proto'
  full_command = command.join ' '

  sh full_command
end
```

The path version works fine, but the github version throws an error because of some files that are `.gitignore`'d.

```bash
$ bundle exec rake
<bunch of output>
loader.rb:17:in `require': cannot load such file -- google/api/annotations.pb (LoadError)
```

Is there a good way to ensure the github version works? Or, perhaps a better way to do this?

blowmage:
AFAIK the only way to allow this is to check the generated `*.pb.rb` files into SCM. This was recently done for the common protos gems, and I think it would make sense for us to do something similar in this repo. That way these files would exist and the gem could be referenced either by path or github project.

Is this a change we should make?

jbolinger:
If that change enables this I'm all for it. Also, if any of those generated protos are exposed and would be considered part of our API for customizations I'd be in favor of doing it for that reason as well.

blowmage:
You can now add the gem by putting the following in a Gemfile:

```ruby
gem "gapic-generator", github: "googleapis/gapic-generator-ruby"
```

PR #45, opened 2019-03-13
Add failing test for demo prep

jbolinger:
I would like to be able to have folks do something along the lines #44 for the demo next week. They should be able to build an API (showcase in particular) using just a simple Gemfile/Rakefile pair that will take care of installing everything beyond Ruby and bundler. We'll also show how to write a simple script to use the generated library (and we'll likely have rake run it after generation so it's a one liner demo) i.e.:

```ruby
require "google/showcase/v1alpha3/echo"

client = Google::Showcase::V1alpha3::EchoClient.new(
  credentials: GRPC::Core::Channel.new(
    "localhost:7469", nil, :this_channel_is_insecure))

response = client.echo 'hi there!', nil

puts response.content
```

However, we need to get this test that I've added to pass in order to do that. I was able to edit the generated files by hand to get this test to pass without too much trouble, so I think we can make this happen in time.

To be clear, I don't want to use this integration test in this PR for the demo directly. We'll create a very simple standalone Ruby project that will be very similar. But, we have to get this test to pass for the demo to actually work.

So, let's try to get this test in quickly and then jump on to the harder part - getting it to pass 😄 .

### DIFF
--- a/shared/Rakefile
+++ b/shared/Rakefile
@@ -73,9 +73,26 @@ end
 namespace :test do
   desc "Run functional tests for showcase protos"
   Rake::TestTask.new :showcase do |t|
+    require "tmpdir"
+
+    client_lib = Dir.mktmpdir
+    protoc_cmd = [
+      "grpc_tools_ruby_protoc",
+      "--proto_path=api-common-protos",
+      "--proto_path=protos/showcase",
+      "--ruby_out=#{client_lib}",
+      "--grpc_out=#{client_lib}",
+      "--ruby_gapic_out=#{client_lib}",
+      "protos/showcase/v1alpha3/echo.proto"
+    ].join " "
+    sh protoc_cmd
+
     t.libs << "test"
+    t.libs << client_lib
     t.test_files = FileList["test/showcase/**/*_test.rb"]
     t.warning = false
+
+    # TODO: delete the temporary dir?
   end
 end
 
@@ -97,6 +114,14 @@ task :console do
   ARGV.clear
   IRB.start
 end
+
+desc "Run the CI build"
+task :ci do
+  puts "\nshared test\n"
+  Rake::Task[:test].invoke
+end
+
+task default: :ci
 
 def generate_input_file name, protos
   require "tmpdir"
--- a/shared/Rakefile
+++ b/shared/Rakefile
@@ -71,25 +71,9 @@ end
 namespace :test do
   desc "Run functional tests for showcase protos"
   Rake::TestTask.new :showcase do |t|
-    require "tmpdir"
-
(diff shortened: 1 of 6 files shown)


blowmage:
> I was able to edit the generated files by hand to get this test to pass

What edits to the generated files are you making?

shared/Rakefile
```diff
@@ -65,17 +65,31 @@ end
...
-  t.test_files = FileList["test/**/*_test.rb"]
-  t.warning = false
+task :test do
+  Rake::Task["test:showcase"].invoke
 end
 namespace :test do
   desc "Run functional tests for showcase protos"
   Rake::TestTask.new :showcase do |t|
-    t.libs << "test"
+    require "tmpdir"
```
blowmage:
```suggestion
    require "minitest"
    require "fileutils"
    require "tmpdir"
```

shared/Rakefile
```diff
@@ -65,17 +65,31 @@ end
...
+      "--ruby_gapic_out=#{client_lib}",
+      "protos/showcase/v1alpha3/echo.proto"
+    ].join " "
+    sh protoc_cmd
+
+    t.libs << ["test",  client_lib]
     t.test_files = FileList["test/showcase/**/*_test.rb"]
     t.warning = false
+
+    # TODO: delete the temporary dir?
```
blowmage:
```suggestion
    Minitest.after_run do
      FileUtils.remove_dir client_lib, true
    end
```

shared/Rakefile
```diff
@@ -65,17 +65,31 @@ end
...
+task :test do
+  Rake::Task["test:showcase"].invoke
 end
 namespace :test do
   desc "Run functional tests for showcase protos"
   Rake::TestTask.new :showcase do |t|
-    t.libs << "test"
+    require "tmpdir"
+
+    client_lib = Dir.mktmpdir
```
blowmage: I think creating the temp directory, running the `protoc` command, and cleaning up the temp directory could all be done in the test helper. The test helper already starts and stops the echo server. I don't think this _needs_ to be performed in the rake task.

jbolinger: Yeah, that's true. I'll move it.

jbolinger:
@blowmage I put the client that I got the test to pass with in #46. Note that it's just an example and there is no reason that we have to do it exactly like that. I just want to get us in a state where the client is runnable.

shared/Rakefile
jbolinger: I added this when I moved it to the helper. Tx!

shared/Rakefile
jbolinger: Added in the helper.

PR #46, opened 2019-03-13
Manually fixed showcase client

jbolinger:
This contains a hand edited showcase client that will pass the test added in #45. It's meant as an example only to show what needs to be fixed in order to generate a functional client.

Note that many of the changes here are arbitrary. I simply removed things that are not working and added some strange workarounds. It's clearly not the ideal output - this is just to get into a state where the output is runnable.

This PR is not meant to be merged.


quartzmo:
Just FYI, on this branch (but not on master) I get:

```sh
Chriss-iMac:gapic-generator-ruby quartzmo$ bundle exec rake update
Running bundle update for shared
bundle update
Could not find gem 'grpc-tools (~> 1.18)' in any of the gem sources listed in your Gemfile.
rake aborted!
Command failed with status (7): [bundle update...]
/Users/quartzmo/code/google/codez/gapic-generator-ruby/Rakefile:115:in `block (3 levels) in <top (required)>'
```

jbolinger:
@quartzmo / @blowmage I'm going to close this. It looks like you're already past this stage and we no longer need this reference. Let me know if that's not the case.

PR #47, opened 2019-03-13
Fix CircleCI Build

blowmage:
The fix is to remove the cache (for now) and use the `grpc-tools` gem to call `protoc`.


.circleci/config.yml
```diff
@@ -6,26 +6,30 @@ version: 2
...
+      - image: circleci/ruby:2.6.1
     working_directory: ~/gapic-generator-ruby
     steps:
       - checkout
+      - run:
+          name: update environment
+          command: |
+            gem update --system
+            gem install bundler -v 1.7.3
+            apt-get install -yqq protobuf-compiler
```
jbolinger: What do you think about standardizing on using the `grpc-tools` gem for this, which comes with protoc?

blowmage: I don't know anything about grpc-tools.

blowmage: If we add that gem as a dev dependency can we remove the "Install the Proto Compiler" section of the gapic-generator README?

jbolinger: The short version is this:

```bash
$ gem install grpc-tools
$ grpc_tools_ruby_protoc --proto_path=my/protos my/protos/foo.proto
```

In other words, this would allow us to keep everything in Ruby land and install protoc via ruby gems. It also won't interfere with with a local protoc since the binary is named differently.

For a full example you can see how I used it in #45.

The only real downside I can see to this approach is them gem lags new protoc releases a bit, but I really doubt that would ever matter.

jbolinger: Yes, that would no longer be needed.

blowmage: The [build with grpc-tools](https://circleci.com/gh/googleapis/gapic-generator-ruby/27) failed, but the [build with apt-get](https://circleci.com/gh/googleapis/gapic-generator-ruby/26) passed.

gapic-generator/gapic-generator.gemspec
```diff
@@ -46,6 +46,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "rubocop", "~> 0.61"
 
   spec.add_development_dependency "bundler", "~> 1.17"
+  spec.add_development_dependency "grpc-tools", "~> 1.19"
```
jbolinger: don't we need this as a runtime dependency?

.circleci/config.yml
jbolinger: hmm. we have to rename any use of `protoc` to `grpc_tools_ruby_protoc` but that error doesn't look like that's the problem.

jbolinger: we likely need to remove stuff like this as well: `--plugin=protoc-gen-ruby-protobuf=/usr/local/bundle/bin/protoc-gen-ruby`, but I'm still confused as to why it would lead to that error.

blowmage: K, removed `apt-get` and switched to using `grpc_tools_ruby_protoc`. The [build passed](https://circleci.com/gh/googleapis/gapic-generator-ruby/28).

gapic-generator/gapic-generator.gemspec
blowmage: Not AFAICT. The generator shouldn't mandate how it is used. Users should be able to use the generator by either installing `protoc` using `apt-get`, or compiling it, or installing `grpc-tools` and calling `grpc_tools_ruby_protoc`.

jbolinger: I suppose that's true. Our first goal is to be a protoc plugin and you're totally right about not mandating in that respect so let's keep it like this.

PR #51, opened 2019-03-15
Update generator for Showcase client WIP

quartzmo:
Add `showcase.yml`
Update gapic-generator client.erb
Update gapic-generator api.rb
Update shared/test_helper.rb

### DIFF
--- a/gapic-generator/templates/default/shared/_client.erb
+++ b/gapic-generator/templates/default/shared/_client.erb
@@ -6,8 +6,20 @@ require "google/gax"
 require "google/gax/operation"
 require "google/longrunning/operations_client"
 
+<% if service.name == "Echo" %>
+require "google/showcase/v1alpha3/echo_pb"
+
+module Google
+  module Gax
+    ::CallSettings = CallSettings
+  end
+end
+<% else %>
 require "google/cloud/<%= service.name.underscore %>/v1/cloud_<%= service.name.underscore %>_pb"
+<% end %>
+<% unless api.configuration[:skip_credentials] %>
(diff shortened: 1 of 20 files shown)


jbolinger:
@quartzmo I think this puts on on a good track for the demo next week. How about doing the minimal amount here to get a green build, so we're in a good state, and then seeing how much more of the fakery we can remove in followups? Perhaps starting with the imports or the `CallSettings` that is being special cased here?

gapic-generator/templates/default/shared/_client.erb
```diff
@@ -97,13 +114,15 @@ class <%= service.name %>Client
...
+<% unless api.configuration[:skip_credentials] %>
     defaults = default_settings client_config, timeout, metadata, lib_name, lib_version
+<% end %>
 
 <% service.methods.each do |method| %>
 <% method_name = method.descriptor.name.underscore %>
 @<%= method_name %> = Google::Gax.create_api_call(
 @<%= service.name.underscore %>_stub.method(:<%= method_name %>),
-defaults["<%= method_name %>"],
+CallSettings.new,
```
jbolinger: Does this line need to be special cased with the `if service.name == "Echo"`? Or, maybe we should just always add that `CallSettings` trick until we do this part for real?

quartzmo: @jbolinger Agreed, we got there by accident, I'll remove that conditional.

blowmage:
@quartzmo You can run `bundle exec rake gen` from the gapic-generator directory to update the expected output files to match the changes in the generator.

PR #52, opened 2019-03-18
Enable and test all supported annotations

blowmage:
This PR enables all the annotations and adds test coverage for them.


landrito:
This is awesome! Will look at it soon!

blowmage:
Rebased on master to bring this up to date.

Right now the Rubocop rules are being applied to the tests, so we either need to disable some rules in the tests, or I need to refactor the tests to pass the Rubocop check.

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -217,11 +217,9 @@ def host
...
         end
 
-        # @return [Google::Api::OAuth] OAuth information for the client.
-        #   The "scopes" key is a repeated string; see
-        #   `google/api/metadata.proto`.
-        def oauth
-          options[:".google.api.default_host"] if options
+        # @return [String] The OAuth scopes information for the client.
+        def scopes
+          options[:".google.api.oauth_scopes"] if options
```
landrito: This is a comma separated string. Maybe it would be best to do the split here and return an array?

jbolinger: Is it worth breaking this up into any `Array` here instead of letting the caller do that?

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -261,26 +259,26 @@ def initialize descriptor, address, docs, input, output
           @output = output
         end
 
-        # @return [Array<Google::Api::MethodSignature>] The parameter lists
+        # @return [Array<String>] The parameter lists
         #   defined for this method. See `google/api/signature.proto`.
```
landrito: `google/api/signature.proto` =>`google/api/client.proto`

gapic-generator/lib/google/gapic/schema/wrappers.rb
```diff
@@ -261,26 +259,26 @@ def initialize descriptor, address, docs, input, output
           @output = output
         end
 
-        # @return [Array<Google::Api::MethodSignature>] The parameter lists
+        # @return [Array<String>] The parameter lists
```
jbolinger: Similar to `scopes`, should this be at the same level of abstraction as the proto or should we break this down so the caller doesn't need to know it's `,` delimitated? (also I think the comment is outdated and `signature.proto` was moved or renamed)

gapic-generator/test/google/gapic/annotations/method_test.rb
```diff
@@ -0,0 +1,242 @@
...
+    method = service.methods.find { |s| s.name == "GetRepeatedGarbage" }
+    refute_nil method
+
+    assert_kind_of Google::Protobuf::MethodOptions, method.options
+    assert_kind_of Google::Api::HttpRule, method.options[:http]
+    assert_equal "/v1/repeated_garbage:get", method.options[:http][:post]
+    assert_equal "*", method.options[:http][:body]
+
+    expected_signatures = [
+      "name,bool,enum",
```
jbolinger: Might not be a bad idea to add some nested signatures in one of these tests. i.e. `name,bool,foo.bar.baz`

gapic-generator/lib/google/gapic/schema/wrappers.rb
blowmage: It is a single string with commas. Updated to return an array of scope strings.

gapic-generator/lib/google/gapic/schema/wrappers.rb
blowmage: Updated.

gapic-generator/lib/google/gapic/schema/wrappers.rb
blowmage: Updated.

gapic-generator/lib/google/gapic/schema/wrappers.rb
blowmage: Updated.

gapic-generator/test/google/gapic/annotations/method_test.rb
blowmage: I don't understand how that would work. How do you define nested parameters to a Ruby method?

blowmage:
> Can you remove `resource_set` stuff for now. It wasn't added in this round of annotations.

Removed.

> References to `google/api/resources.proto` should be changed to `google/api/resource.proto` (singluar)

Updated.

> Nit: Maybe we can add documentation to either the test or the garbage proto to document what is being covered.

I consider the tests to be documentation of what is being covered. What type of documentation are you thinking of?

gapic-generator/test/google/gapic/annotations/method_test.rb
landrito: Meaning a nested field in a message.

example:

```proto
rpc GetThing(ThingRequest) returns (ThingResponse) {
    option (google.api.method_signature) = "top_level_field,submessage.nested_field";
}

message ThingRequest {
  string top_level_field = 1;
  Submessage submessage = 2;
}

message Submessage {
  string nested_field = 1;
}
```

this should yield a method whose signature can be called like:

```rb
client.get_thing(top_level_field: "top", nested_field: "nested")

# or this depending on how we want to handle some disambiguation cases

client.get_thing(top_level_field: "top", submessage_nested_field: "nested")

Comment (blowmage) [reply gapic-generator/test/google/gapic/annotations/method_test.rb]: In that case the signature would be `top_level_field,submessage`, wouldn't it? The way you would call it is:

```ruby
client.get_thing(top_level_field: "top", submessage: { nested_field: "nested" })
```

We've talked about not even using these signatures outside of possibly documentation code examples. I don't know how we would support dotted field names in ruby method arguments, as Ruby does not have method overloading like other languages, but there is no reason dotted field names can't exist in the annotations.

landrito:
On second look at the test I do agree that it's self documenting enough.

gapic-generator/test/google/gapic/annotations/method_test.rb
landrito: >In that case the signature would be top_level_field,submessage, wouldn't it? 
Nope the new annotations will allow for `top_level_field,submessage.nested_field`

Can you give an example of how dotted names couldn't work?

blowmage: Can we table this for now and address it in a separate issue?

landrito: Let's table how we support nested fields but can you add the nested field to the annotation since the it is a valid input.

blowmage: You want me to add an annotation for a nested field even though we don't have any nested fields to reference? So a method signature of `name,foo.bar.baz`?

blowmage: Added.

landrito: Ah sorry I was unclear. To be more specific. Heres the changes:

```diff
rpc GetSpecificGarbage(SpecificGarbage) returns (SpecificGarbage) {
  ...
  option (google.api.method_signature) = "name,int32,bool";
+ option (google.api.method_signature) = "name,nested.int64";
}

message SpecificGarbage {
  string name = 1 [(google.api.field_behavior) = REQUIRED];
  ...
  GarbageEnum enum = 11;
+ NestedGarbage nested = 12;

  message NestedGarbage {
    ...
    int64 int64 = 3;
    ...
  }
}
```

blowmage: Added.

PR #54, opened 2019-03-19
Add YAML input files

blowmage:
This makes it easier to view the contents of the proto request. Useful for debugging annotations and whatnot, as you can compare the diffs to the YAML files.


blowmage:
This change isn't strictly neccessary. It has been valuable to examine the changes as I've been trying to enable the annotations, so I figured it was worth posting it so you can weigh in on whether it is useful to keep around or something to throw away.

landrito:
Do we have to check in these yaml files? If so can we append `descriptor` or `desc` to the yaml file name?

blowmage:
> Do we have to check in these yaml files?

The value I got from them is checking them in, and then changing the proto compile settings and doing a `git diff` to see what changed. So I do think there is value in having these files checked in.

> If so can we append descriptor or desc to the yaml file name?

We can call it whatever we want. It is the YAML version of the contents of the bin file, so to me it makes sense that the files would use the same name except for the different extension, which indicates the format. Why would you want the YAML files to have a different file name?

landrito:
In that case can we change both the bin and the yaml to be `{api}_desc.{ext}` I am asking because I would like these yamls to be distinct from a configuration yaml file that we will be expecting as input.

blowmage:
> I am asking because I would like these yamls to be distinct from a configuration yaml file that we will be expecting as input.

Changed.

PR #56, opened 2019-03-20
De-fake the cloud generator

blowmage:
Remove all GCP specific settings and references. Missing annotations can be provided using the configuration file.


blowmage:
I can see restoring the prior credentials file, and considering it something that belongs at the package level. And the scopes would be a combination of all services defined at that package, e.g. `google.cloud.speech.v1`. Do we want to make that change in this PR, or a separate PR?

blowmage:
Per @landrito's offline request, I have removed the expected output for all four configured services from the cloud generator tests, and the change to the showcase test to use the cloud generator. PTAL.

gapic-generator-cloud/templates/cloud/client/_client.erb
```diff
@@ -81,32 +98,20 @@ class <%= client_name service %>
 
   def default_settings client_config, timeout, metadata, lib_name,
                         lib_version
-    package_version = Gem.loaded_specs["<%= gem_name api %>"]
-      .version.version
+    package_gem = Gem.loaded_specs["<%= gem_name api %>"]
+    package_version = package_gem ? package_gem.version.version : nil
 
     google_api_client = ["gl-ruby/#{RUBY_VERSION}"]
```
landrito: Bah I originally wrote this and it looks ugly. Can you make this look nicer by using `.join " "`

gapic-generator-cloud/templates/cloud/helpers/cloud_helper.rb
```diff
@@ -50,47 +88,145 @@ def client_lro? _service
...
-    ["https://www.googleapis.com/auth/cloud-platform"]
+  def helpers_file_path service
+    helpers_require(service) + ".rb"
   end
 
-  def service_proto_require _service
-    "google/cloud/speech/v1/cloud_speech_services_pb"
+  def client_address service
+    address = service.host
+    address ||= service.parent.parent.protoc_options[:default_host]
```
landrito: For options that map to a proto annotation option, let's not expose a protoc/configuration interface since we want to be strict about where that information is coming from. If we get user need in the future we can reevaluate then.

blowmage: Do we want to look at the protoc options for anything other than generator type and configuration path then? Should we remove the check against `protoc_options` from everywhere?

gapic-generator-cloud/templates/cloud/client/_client.erb
blowmage: Done.

gapic-generator-cloud/templates/cloud/helpers/cloud_helper.rb
quartzmo: I agree that we should be strict about where information is coming from, it will make maintenance much easier. I would be in favor of removing the check against protoc_options from everywhere.

landrito: Also in favor of removing protoc_options check from everywhere.

blowmage:
All issues have been created.

gapic-generator-cloud/templates/cloud/helpers/cloud_helper.rb
blowmage: Moving discussion to #71.

PR #73, opened 2019-03-20
Remove YAML input files

blowmage:
Different platforms are generating slightly different files.
So the files are showing changes even when there are no
substantive changes. Better to remove these files then.


quartzmo:
Hold off on merging this please until we see if somehow we can all get the same output.

quartzmo:
It seems to me that we should at least keep `generate_yaml_file` as a separate rake task.

blowmage:
Generating the YAML file is super easy. Open a console and do the following:

```
cd gapic-generator
bundle exec rake console
```

And type the following:

```ruby
speech = schema :speech
require "yaml"
File.write "speech.yaml", YAML.dump(speech.request.to_hash)
```

I don't know if that is worth having a rake task that does it and puts the file where git will ignore it. This feature is nice if it adds zero friction. I think if there is any friction we should remove it.

quartzmo:
OK. Can you describe here the way you used the feature, to preserve the technique?

blowmage:
While I was making changes to the generator to get annotations working, I would generate a YAML file and commit it. Then I would make the changes and generate the YAML again. Then `git diff` would show what changes were made. But you could just as easily save the before and after to two different files and do a diff between them.

PR #75, opened 2019-03-21
Remove usages of protoc_options

### LINKED ISSUE #71: Do not look in configuration file for annotation values
We should only use `protoc_options` for very high level options, not for specifying package-level configuration. For package-level values we should only look at the annotations and the configuration file (that was specified in the `protoc_options`).

quartzmo:
[fixes #71]

### DIFF
--- a/gapic-generator-cloud/templates/cloud/helpers/cloud_helper.rb
+++ b/gapic-generator-cloud/templates/cloud/helpers/cloud_helper.rb
@@ -18,9 +18,7 @@ require "active_support/inflector"
 
 module CloudHelper
   def gem_name api
-    api.protoc_options[:gem_name] ||
-      api.configuration[:gem_name] ||
-      gem_address(api).join("-")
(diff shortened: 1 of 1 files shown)


quartzmo:
I would also like to remove [`Api#protoc_options`](https://github.com/googleapis/gapic-generator-ruby/blob/af324f78b6a9b1f7fdd0fefeebe13b66191449b4/gapic-generator/lib/google/gapic/schema/api.rb#L59) (at least make it `protected`), but it is used at [default_generator.rb#L36](https://github.com/googleapis/gapic-generator-ruby/blob/af324f78b6a9b1f7fdd0fefeebe13b66191449b4/gapic-generator/lib/google/gapic/generators/default_generator.rb#L36) for the configuration of a custom path for the templates directory. Can this be changed?

blowmage:
I don’t think removing or protecting `protoc_options` is necessary.

Issue #58, opened 2019-03-20
maintain backwards compatibility with existing client class (SpeeechClient = Speech::Client)


blowmage:
PR #56 moved the client class to be at the service level (Google::Cloud::Speech::V1::Speech::Client`) rather than at the package level (Google::Cloud::Speech::V1::SpeechClient`). Do we want to move the client class back to the previous location? It is possible to alias the old client constant to the new client constant. Do we want to keep it where it is? Move it back? Or add an alias?

blowmage:
Decided not to support this.

PR #77, opened 2019-03-22
Add service .new constructor

blowmage:


blowmage:
I think this is related to #58. Ernest told us that some folks may dislike the duplication of "Speech" in the following:

```ruby
Google::Cloud::Speech::V1::Speech::Client.new
```

The current generator solves this by appending "Client: to the service:

```ruby
Google::Cloud::Speech::V1::SpeechClient.new
```

But this change allows you to create a client object from the service module:

```ruby
Google::Cloud::Speech::V1::Speech.new
```

The service module is where the `Service` class, the `Stub` constant, the `Client` class, and the `Credentials` class are all currently defined. I think there is some elegance to keeping them all in the service module, and using a `.new` method to get a client object.

In fact, I think the service module is the only place a `.new` method makes sense. Not at the version module (`Google::Cloud::Speech::V1.new`), and not at the package level (`Google::Cloud::Speech.new`).

jbolinger:
Ah, I see. I agree that moving everything under the service module (i.e. `Google::Cloud::Speech::V1::Speech`) looks great and this change also makes the client match the fully qualified proto service path verbatim (which I also think is a big plus).

But, this doesn't remove the duplication and this doesn't make the clients backwards compatible (per #58). So, I'm in favor of this but I think we should resolve that issue first and fully commit to the service module concept. I don't see any reason to maintain backwards compatibility given the method signature changes in the new clients and I don't see any problems with the duplication since it's using the path + name without changes. 

Other opinions?

blowmage:
> I don't see any reason to maintain backwards compatibility given the method signature changes in the new clients

FWIW, I agree. We can always add backwards compatibility shims if that is really wanted.

blowmage:
> I don't see any problems with the duplication since it's using the path + name without changes.

In the original showcase echo test that @quartzmo added, the line that created the client object was:

```ruby
client = Google::Showcase::V1alpha3::Echo.new
```

When the showcase echo test was converted to the cloud generator I changed the line to:

```ruby
client = Google::Showcase::V1alpha3::Echo::Client.new
```

This change allows it to go back to what it was. And it make sense IMO. A `google/showcase/v1alpha3/echo.proto` file that defines an `Echo` service currently generates the following files:

* `google/showcase/v1alpha3/echo_pb.rb`
* `google/showcase/v1alpha3/echo_service_pb.rb`
* `google/showcase/v1alpha3/echo_client.rb`

I think it would be an improvement to generate this instead:

* `google/showcase/v1alpha3/echo_pb.rb`
* `google/showcase/v1alpha3/echo_service_pb.rb`
* `google/showcase/v1alpha3/echo.rb`

jbolinger:
I'm happy to drop the `_client` suffix, and that'll make it more consistent with this change.

blowmage:
k, I'll update this PR to change the filename as well.

blowmage:
Updated. Honestly, I like this change more than I thought I would. PTAL.

jbolinger:
I like it. Tests are failing though so I think you missed a reference or two.

quartzmo:
@blowmage What is the type returned by `Google::Cloud::Speech::V1::Speech.new`?

blowmage:
> What is the type returned by `Google::Cloud::Speech::V1::Speech.new`?

It will return a `Google::Cloud::Speech::V1::Speech::Client` object.

blowmage:
Oops. Yeah, the showcase test wasn't updated to use the new client file path. Fixed now. 🤞

Issue #59, opened 2019-03-20
Add env_prefix value to the configuration


blowmage:
One of the settings that is currently missing in the cloud generator is a product_name that can be used as the prefix to the Credentials' environment variables.

We should allow this value to be configured in the configuration file. It might also make sense to infer a product name prefix from the gem_name. A prefix for the gem `google-cloud-speech` is `SPEECH`, so an appropriate fallback would be `gem_name.split("-").last`.

jbolinger:
Do you think using the annotations for `title` or `product_title` would be okay to use in cases where the name has to be inferred?

blowmage:
Here is the example from the `client.proto` file:

```protobuf
option (google.api.package) = {
  title: "Pub/Sub"
  namespace: ["Google", "Cloud"]
  version: "v1"
};
```

I don't know how any of these values are intended to be used, but I specifically don't understand how "Pub/Sub" is to be used in our context. We can't really use that value in the Ruby namespace, and some other examples show spaces in the title. My concern is that the format of the title is not a great fit for how we want to use the title in the environment variables. So, I think we should also allow for a product name that fits our usage to be configured in the configuration file.

jbolinger:
To clarify, I think having whatever configuration we want is fine to allow for fine tuning. Something like a gem name doesn't have a clear mapping to any these options and there's no reason to force it, and I think a config file for that is a great way to go about it.

However, I do think we could use one of these as a default value so that we do not have to fail if no option is specified and `title` seems like a good one to me (of course, we'd have to normalize the string to be valid Ruby).

The meaning of each option is well documented but there's some interpretation needed for how they translate to Ruby. We should probably open separate issues for cases where it's not clear and needs discussion. I think what you're calling `product_name` here equivalent to `title` or `product_title` depending on the context.

blowmage:
The other reason I didn't initially use the service title annotation is that it is not specific to the gem. For gems that have many packages, and each package having many services, we run into an issue of which service to take the value from.

In the end, the only use for this value right now is the environment variable prefix. So I am going to switch this issue to be specific to that. If we every do need a gem-wide title we should open a new issue.

PR #78, opened 2019-03-23
Add Package

### LINKED ISSUE #67: Fix Rakefile Credentials references
How does the Rakefile know which API/package Credentials file to require and use? Is this configurable? Is it somehow infered?

blowmage:
This PR adds the concept of Package. The relationship is:

* API
  * Package
    * Service
      * Method
        * Field

A Package can also be considered the API version. A gem can be generated with several packages, e.g. `google.cloud.speech.v1` and `google.cloud.speech.v1beta1`. Each package/version has a file that can be required, which will in turn require all services defined in the package/version.

With this change new presenters have been added, and most helper methods have been moved to the respective presenter. New helper method have been added to create presenters, and these presenter helper methods can be included in the generator, so that the generator can also use the presenters.

All the cloud templates have been updated to use the presenter objects.

Minimal changes have been made to the Schema classes in this PR.

[refs #59]
[fixes #67]


blowmage:
I have updated this PR to resolve the merge conflict created by merging #77.

PR #81, opened 2019-03-25
Add ServicePresenter#stub_name

### LINKED ISSUE #61: Rename Client's stub ivar?
PR #56 renamed the client's `stub` ivar to be `@stub`. But there are reasons why this stub name was customized. Are these conscerns still valid on the new generator? Should the stub ivar be renamed to something different on each client?

blowmage:
[closes #61]


blowmage:
@landrito does this look right to you? I'll wait on an approval from you on this.

Issue #69, opened 2019-03-20
Update Client method signatures


blowmage:
When calling a streaming method, an enum of request objects is sent in a single method argument. But when calling a non-streaming method the individual fields of the request object are used as the method arguments. We would like to revise how client methods are defined.

We have a couple ideas. One is to change the request object's individual fields to all be named arguments instead of positional arguments. We would also like to allow passing in a request object instead of the individual field values, either as a proto resource object or a Ruby Hash.

blowmage:
We decided this was resolved by #82.

PR #82, opened 2019-03-25
Update client method signatures

blowmage:
* Allow request objects to be passed as well as named arguments.
* Individual request field values must now used named arguments.

[refs #69]

### DIFF
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
@@ -2,16 +2,32 @@
 ##
 <%- if method.doc_description -%>
 <%= indent method.doc_description, "# " %>
-<%- end -%>
 #
-<%- method.fields.each do |field| -%>
-# @param <%= field.name %> [<%= field.doc_types %>]
-<%- if field.doc_description -%>
-<%= indent field.doc_description, "#   " %>
+<%- end -%>
+# @overload <%= method.name %>(request, options: nil)
+#   @param request [<%= method.request_type %> | Hash]
+<%- if method.doc_description -%>
+<%= indent method.doc_description, "#     " %>
+<%- end -%>
+#   @param options [Google::Gax::CallOptions]
+#     Overrides the default settings for this call, e.g, timeout, retries, etc.
+#
+<%- arg_list = method.arguments.map { |arg| "#{arg.name}: nil"}.join ", " -%>
+# @overload client_method(<%= arg_list %>, options: nil)
+<%- method.arguments.each do |arg| -%>
+#   @param <%= arg.name %> [<%= arg.doc_types %>]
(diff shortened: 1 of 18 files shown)


gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
```diff
@@ -2,16 +2,32 @@
...
 <%- end -%>
+#   @param options [Google::Gax::CallOptions]
+#     Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
-<%- method.fields.each do |field| -%>
-# @param <%= field.name %> [<%= field.doc_types %>]
-<%- if field.doc_description -%>
-<%= indent field.doc_description, "#   " %>
+<%- arg_list = method.arguments.map { |arg| "#{arg.name}: nil"}.join ", " -%>
+# @overload client_method(<%= arg_list %>, options: nil)
```
landrito: This needs to be wrapped in erb brackets.

gapic-generator-cloud/templates/cloud/client/method/def/_normal.erb
```diff
@@ -2,16 +2,32 @@
...
 <%- end -%>
+#   @param options [Google::Gax::CallOptions]
+#     Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
-<%- method.fields.each do |field| -%>
-# @param <%= field.name %> [<%= field.doc_types %>]
-<%- if field.doc_description -%>
-<%= indent field.doc_description, "#   " %>
+<%- arg_list = method.arguments.map { |arg| "#{arg.name}: nil"}.join ", " -%>
+# @overload client_method(<%= arg_list %>, options: nil)
```
landrito: Similarly client_method should be wrapped in erb tags.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
landrito: Unless I am misunderstanding `@overload`, client_method should be <%= method.name %>.

blowmage: LOL! Yeah, that is a mistake. That is something from my cheat sheet when I was implementing this. It needs to be `<%= method.name %>`. Fixing now.

blowmage: Fixed.

gapic-generator-cloud/templates/cloud/client/method/def/_normal.erb
blowmage: Fixed.

blowmage:
The output files don't have any changes, CI is passing. The gem name's last node is what is being used right now and it is correct for all the gems. But once we get the garbage proto added to cloud we will use the `env_prefix` configuration.

PR #83, opened 2019-03-25
Add client alias

blowmage:
This alias maintains backwards compatibility with the current gapic generator output.


blowmage:
This PR is an attempt to fix #58. However, not that #77 has been accepted I feel this is a step backwards. I think that this can be done by clients using Autosynth, for those that want this backwards compatibility. But, I think this is a step backwards. What do you all think?

jbolinger:
I don't think there is any reason to do this given the other breaking changes. My vote is to drop it and close #58.

blowmage:
K, will gladly close!

PR #84, opened 2019-03-25
Refactor product_env_prefix

### LINKED ISSUE #59: Add env_prefix value to the configuration

blowmage:
* Lookup value from configuration.
* Fallback by inferring prefix from gem_name.

[closes #59]

### DIFF
--- a/gapic-generator-cloud/templates/cloud/helpers/presenters/api_presenter.rb
+++ b/gapic-generator-cloud/templates/cloud/helpers/presenters/api_presenter.rb
@@ -78,11 +78,7 @@ class ApiPresenter
       "https://github.com/googleapis/googleapis"
   end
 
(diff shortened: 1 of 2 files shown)


blowmage:
The output files don't have any changes as CI is passing. The gem name's last node is what is being used right now and it is correct for all the gems. But once we get the garbage proto added to cloud we will use the `env_prefix` configuration.

PR #86, opened 2019-03-26
Refactor base generator code formatting

blowmage:
Move the formatter into the base generator.
Rename the generate_file method, add convenience name as alias.
Add config method that allows plugin generators to override
the location of the configuration file to use to format.
Add formatter code as separate method so it can also be
overridden. (Hi, rubyfmt)
Fix the formatter so it works in the cloud generator.

### DIFF
--- a/gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
+++ b/gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
@@ -49,15 +49,15 @@ module Google
           ap.packages.each do |package|
             # Package level files
             files << g("package.erb", "lib/#{package.version_file_path}",
-                       package: package, format: true)
+                       package: package)
 
             package.services.each do |service|
               # Service level files
               files << g("client.erb", "lib/#{service.client_file_path}",
-                         service: service, format: true)
+                         service: service)
               files << g("client_test.erb",
                          "test/#{service.client_test_file_path}",
-                         service: service, format: true)
+                         service: service)
             end
           end
 
--- a/gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
+++ b/gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
@@ -48,28 +48,36 @@ module Google
 
           ap.packages.each do |package|
             # Package level files
-            files << cop(gen("package.erb", "lib/#{package.version_file_path}",
-                             package: package))
+            files << g("package.erb", "lib/#{package.version_file_path}",
+                       package: package, format: true)
 
             package.services.each do |service|
               # Service level files
-              files << cop(gen("client.erb",
-                               "lib/#{service.client_file_path}",
-                               service: service))
-              files << cop(gen("client_test.erb",
-                               "test/#{service.client_test_file_path}",
-                               service: service))
+              files << g("client.erb", "lib/#{service.client_file_path}",
+                         service: service, format: true)
+              files << g("client_test.erb",
+                         "test/#{service.client_test_file_path}",
+                         service: service, format: true)
             end
           end
 
           # Api level files
-          files << gen("gemspec.erb",  "#{ap.gem_name}.gemspec", api: ap)
-          files << gen("gemfile.erb",  "Gemfile",                api: ap)
-          files << gen("rakefile.erb", "Rakefile",               api: ap)
-          files << gen("rubocop.erb",  ".rubocop",               api: ap)
-          files << gen("license.erb",  "LICENSE.md",             api: ap)
+          files << g("gemspec.erb",  "#{ap.gem_name}.gemspec", api: ap)
+          files << g("gemfile.erb",  "Gemfile",                api: ap)
+          files << g("rakefile.erb", "Rakefile",               api: ap)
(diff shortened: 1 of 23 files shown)


gapic-generator/lib/google/gapic/generators/base_generator.rb
```diff
@@ -69,16 +69,33 @@ def controller
           @controller ||= Class.new(ActionController::Base).new
         end
 
-        def gen template, filename, **args
+        def generate_file template, filename, format: true, **args
```
quartzmo: I really like the addition of this `format` option.

gapic-generator/lib/google/gapic/generators/default_generator.rb
```diff
@@ -49,36 +49,12 @@ def generate
           files = []
 
           api_services(@api).each do |service|
-            files << cop(gen("client.erb", ruby_file_path(service),
-                             api: @api, service: service))
+            files << g("client.erb", ruby_file_path(service),
+                       api: @api, service: service, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

shared/output/gapic/gems/my_plugin/lib/google/gapic/generators/my_plugin_generator.rb
```diff
@@ -48,8 +48,8 @@ def generate
           files = []
 
           api_services(@api).each do |service|
-            files << cop(gen("client.erb", ruby_file_path(service),
-                             api: @api, service: service))
+            files << g("client.erb", ruby_file_path(service),
+                       api: @api, service: service, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
```diff
@@ -48,29 +48,37 @@ def generate
 
           ap.packages.each do |package|
             # Package level files
-            files << cop(gen("package.erb", "lib/#{package.version_file_path}",
-                             package: package))
+            files << g("package.erb", "lib/#{package.version_file_path}",
+                       package: package, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
```diff
@@ -48,29 +48,37 @@ def generate
...
             package.services.each do |service|
               # Service level files
-              files << cop(gen("client.erb",
-                               "lib/#{service.client_file_path}",
-                               service: service))
-              files << cop(gen("client_test.erb",
-                               "test/#{service.client_test_file_path}",
-                               service: service))
+              files << g("client.erb", "lib/#{service.client_file_path}",
+                         service: service, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

gapic-generator-cloud/lib/google/gapic/generators/cloud_generator.rb
```diff
@@ -48,29 +48,37 @@ def generate
...
+              files << g("client_test.erb",
+                         "test/#{service.client_test_file_path}",
+                         service: service, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

gapic-generator/gem_templates/generator.erb
```diff
@@ -35,8 +35,8 @@ module Google
           files = []
 
           api_services(@api).each do |service|
-            files << cop(gen("client.erb", ruby_file_path(service),
-                             api: @api, service: service))
+            files << g("client.erb", ruby_file_path(service),
+                       api: @api, service: service, format: true)
```
quartzmo: Why is `format: true` passed here? Isn't `true` the default?

gapic-generator/lib/google/gapic/generators/default_generator.rb
blowmage: No, it shouldn’t be the default

blowmage: @quartzmo I'm rethinking this. Perhaps formatting the output should be default, and users should pass `format: false` to opt-out. Thoughts?

quartzmo: Isn't the method [defined here](https://github.com/googleapis/gapic-generator-ruby/blob/9c0a8a3d420925b60ad136019844f672f1a1a224/gapic-generator/lib/google/gapic/generators/base_generator.rb#L72) with `format: true`?

quartzmo: Yes, I think formatting the output should be default, and users should pass `format: false` to opt-out.

blowmage: kk, i'll change it.

gapic-generator/lib/google/gapic/generators/base_generator.rb
blowmage: This change adds the private method `format_file` that can be overridden if users don't want to use Rubocop to perform the formatting.

PR #91, opened 2019-03-27
Rename GemPresenter

blowmage:
Use "gem" to indicate that we are building a ruby gem package.

### DIFF
--- a/gapic-generator-cloud/templates/cloud/helpers/presenters/method_presenter.rb
+++ b/gapic-generator-cloud/templates/cloud/helpers/presenters/method_presenter.rb
@@ -23,13 +23,13 @@ class MethodPresenter
     @method  = method
   end
 
-  def api
-    service.api
-  end
-
-  def package
-    service.package
-  end
+  # def api
+  #   service.api
+  # end
+  #
+  # def package
+  #   service.package
+  # end
 
   def service
     ServicePresenter.new @method.parent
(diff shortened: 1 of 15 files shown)


blowmage:
FWIW, I think this change makes the templates much easier to read. I never really knew what an `api` presenter was, other than a wrapper for the Schema Api object. But since we were hiding that fact from the templates, stuck out to me. I think it makes more sense from the generator's and the templates' perspective to navigate a gem than an api.

quartzmo:
I see your point, especially with removing the `gem_` prefix from all those methods.

gapic-generator-cloud/templates/cloud/helpers/presenters/method_presenter.rb
```diff
@@ -23,13 +23,13 @@ def initialize method
...
   end
 
-  def api
-    service.api
-  end
-
-  def package
-    service.package
-  end
+  # def api
```
landrito: Why is this chunk of code commented out?

blowmage: It should have been deleted. It exposes schema objects and we don't want that. I actually thought I did delete it...

blowmage: Yep, left a commit unpushed. Removed.

blowmage:
Can I get an approval so I can merge?

PR #94, opened 2019-03-29
Streaming Enhancements

blowmage:
This PR is a draft, and here to collect feedback. We have discussed what changes we might want to see in the future versions of GAX, and this PR includes two that I think would be very useful.

The first is use of a `StreamInput` class that would hold a stream open until `#close` is called on it. We currently have this functionality in the GCP gems, but they are part of the private API and not something that is documented for users.

The second is to support change is to support blocks on streaming server and bidi methods, and to have the block process responses as they stream from the server. This would make the majority of streaming responses easier to consume. There are changes that are needed to GAX to support this.

Thoughts?

### DIFF
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -9,15 +9,19 @@
 # @param options [Google::Gax::CallOptions]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
-# @return [Enumerable<<%= method.return_type %>>]
-#   An enumerable of {<%= method.return_type %>} instances.
+# @yield [response] Called on each streaming responses, when provided.
+# @yieldparam response [<%= method.return_type %>]
+#
+# @return [Enumerable<<%= method.return_type %>, nil>]
+#   An enumerable of {<%= method.return_type %>} instances when a block is not provided.
+#   When a block is provided `nil` is returned.
 #
 # @raise [Google::Gax::GaxError] if the RPC is aborted.
 #
 # @example
 <%= indent method.code_example, "#   " %>
 #
-def <%= method.name %> requests, options: nil
+def <%= method.name %> requests, options: nil, &block
   unless requests.is_a? Enumerable
     raise ArgumentError, "requests must be an Enumerable"
   end
@@ -26,5 +30,5 @@ def <%= method.name %> requests, options: nil
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options)
+  <%= method.ivar %>.call(requests, options, &block)
 end
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -6,7 +6,7 @@
 <%- end -%>
 # @param requests [Google::Gax::StreamInput, Enumerable<<%= method.request_type %> | Hash>]
 #   An enumerable of {<%= method.request_type %>} instances.
-# @param options [Google::Gax::CallOptions]
+# @param options [Google::Gax::ApiCall::Options, Hash]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
 # @yield [response] Called on each streaming responses, when provided.
@@ -31,8 +31,18 @@ def <%= method.name %> requests, options: nil, &block
   end
 
   requests = requests.lazy.map do |request|
+    # Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options, enum_proc: block)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
+  options.merge timeout: @timeout, metadata: metadata, retry_policy: retry_policy
+
+  <%= method.ivar %> ||= Google::Gax::ApiCall.new @<%= method.service.stub_name %>.method :<%= method.name %>
+  <%= method.ivar %>.call(requests, options: options, on_stream: block)
 end
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -4,7 +4,7 @@
 <%= indent method.doc_description, "# " %>
 #
 <%- end -%>
-# @param requests [Enumerable<<%= method.request_type %> | Hash>]
+# @param requests [Google::Gax::StreamInput, Enumerable<<%= method.request_type %> | Hash>]
 #   An enumerable of {<%= method.request_type %>} instances.
 # @param options [Google::Gax::CallOptions]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
@@ -23,7 +23,11 @@
 #
 def <%= method.name %> requests, options: nil, &block
   unless requests.is_a? Enumerable
-    raise ArgumentError, "requests must be an Enumerable"
+    if requests.respond_to? :to_enum
+      requests = requests.to_enum
+    else
+      raise ArgumentError, "requests must be an Enumerable"
+    end
   end
 
   requests = requests.lazy.map do |request|
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -40,7 +40,7 @@ def <%= method.name %> requests, options: nil, &block
   header_params = {} # { name: request.name }
   request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
   metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
-  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
+  retry_policy = {} # retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
   options.merge timeout: @timeout, metadata: metadata, retry_policy: retry_policy
 
   <%= method.ivar %> ||= Google::Gax::ApiCall.new @<%= method.service.stub_name %>.method :<%= method.name %>
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -34,5 +34,5 @@ def <%= method.name %> requests, options: nil, &block
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options, &block)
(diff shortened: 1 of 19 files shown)


shared/test/showcase/echo_test.rb
```diff
@@ -20,11 +20,100 @@
...
+
+    25.times do
+      break if pull_count >= 20
+
+      sleep 1
+    end
+
+    stream_input.close
+
+    assert pull_count >= 20, "should have pulled 20 messages by now"
```
jbolinger: Can this be tightened up so we're not depending on random sleeps (count down latch or similar?)

shared/test/showcase/echo_test.rb
```diff
@@ -20,11 +20,100 @@
...
+      credentials: GRPC::Core::Channel.new("localhost:7469", nil, :this_channel_is_insecure)
+    )
 
     response = client.echo content: 'hi there!'
 
     assert_equal 'hi there!', response.content
   end
+
+  def test_chat_closure
+    client = Google::Showcase::V1alpha3::Echo.new(
```
jbolinger: sort of unrelated nit but can this move to a before/after (also need to shutdown the channel after each test)?

gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
```diff
@@ -4,27 +4,35 @@
 <%= indent method.doc_description, "# " %>
 #
 <%- end -%>
-# @param requests [Enumerable<<%= method.request_type %> | Hash>]
+# @param requests [Google::Gax::StreamInput, Enumerable<<%= method.request_type %> | Hash>]
```
jbolinger: Could you add a test for the other type(s)?

shared/test/showcase/echo_test.rb
```diff
@@ -20,11 +20,100 @@
...
+      end
+    end
+
+    25.times do
+      break if pull_count >= 20
+
+      sleep 1
+    end
+
+    stream_input.close
```
jbolinger: What will happen if methods are called on `stream_input` after the rpc has terminated (i.e. if it errors or aborts)? Would it be useful to have a way to check if the call is active?

shared/test/showcase/echo_test.rb
blowmage: Yes. This PR is a draft because it is quick and dirty to not final.

shared/test/showcase/echo_test.rb
blowmage: Yes. Each test is run on a separate object and will have its own channel.

gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
blowmage: Yes.

shared/test/showcase/echo_test.rb
blowmage: I don't know how `StreamInput` would be able to determine the health of an RPC call using it's enumerable.

shared/test/showcase/echo_test.rb
jbolinger: Oh sorry, didn't notice that. I'll resolve a few of these then until you're at that point.

blowmage:
These changes are dependent on some changes to google-gax. The streaming changes are probably breaking, which is why the branch is named [experimental-2.0](https://github.com/googleapis/gax-ruby/compare/master...blowmage:experimental-2.0). I think this PR is a good place to discuss these changes though. Thoughts?

shared/test/showcase/echo_test.rb
jbolinger: Before we talk about how, do you think it's needed? 

The use case I'm thinking of is something like this:

```ruby
until stream_input.closed?
  stream_input.push # next item
end
```

but this could not depend soley on the user `close`ing it without being confusing IMHO.

blowmage: Calling `#close` is no guarantee that the client has pulled all the requests from the input and closed the stream. We can track whether `#close` has been called, and perhaps not allow more requests to be pushed to it, but we cannot determine the state of the stream from the input object.

jbolinger: Right, that's really my point though. This interface does not provide a way to monitor the lifecycle of a streaming call. Is this input helper enough or should we also provide a way to do that?

blowmage: I don't know how to track the health of the streams, much less how to design an intuitive API to expose it. AFAIK, as long as the enums are active the stream is active.

jbolinger: Well, we could figure that out :), but let's just table this for now. I'd like to see what others think before getting caught up on this (which I'm sort of importing from what's been done elsewhere and may not be a fit here).

blowmage:
This PR has been updated with all of the changes that have been made in master so far, and it is targeting the [experimental-2.0.1](https://github.com/googleapis/gax-ruby/compare/master...blowmage:experimental-2.0.1) branch that incorporates all the changes made to the Gax 2.0 client, and some other changes that have not yet been added. It shows how responsibility can be moved from Gax to the GAPIC client.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
```diff
@@ -22,12 +22,26 @@
...
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options, &block)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: `UNAVAILABLE` cannot be a default.

Unfortunately, we will probably need a way to configure this in practice because it is correct in many cases. However, there is no configuration flag for it in the spec so we'll have to come up with one, likely via the yaml file you introduced. I'd suggest removing this as a default for now and we can decide how to allow a user to configure it in a later PR.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
```diff
@@ -46,13 +47,19 @@ def <%= method.name %> request = nil, options: nil, **request_fields
...
-    call_options: options
-  )
-  operation.on_done { |operation| yield operation } if block_given?
-  operation
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: This may be okay for LRO, but I'm not sure so let's keep it consistent and remove it for now.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
```diff
@@ -22,12 +22,26 @@
...
 
   requests = requests.lazy.map do |request|
+    # Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options, &block)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
```
jbolinger: Still a WIP?

gapic-generator-cloud/templates/cloud/client/method/def/_server.erb
```diff
@@ -46,7 +50,17 @@ def <%= method.name %> request = nil, options: nil, **request_fields
...
+  # request = Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
   request = Google::Gax.to_proto request, <%= method.request_type %>
 
-  <%= method.ivar %>.call(request, options)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: Can't be a default.

gapic-generator-cloud/templates/cloud/client/method/def/_paged.erb
```diff
@@ -0,0 +1,65 @@
...
+  request ||= request_fields
+  # request = Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
+  request = Google::Gax.to_proto request, <%= method.request_type %>
+
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: Can't be a default.

shared/output/cloud/speech/lib/google/cloud/speech/v1/speech/client.rb
```diff
@@ -80,26 +80,17 @@ class OperationsClient < Google::Longrunning::OperationsClient
...
-            #   An optional proc that intercepts any exceptions raised during an API call to
-            #   inject custom error handling.
             #
             def initialize \
                 credentials: nil,
                 scopes: ALL_SCOPES,
-                client_config: {},
                 timeout: DEFAULT_TIMEOUT,
                 metadata: nil,
-                exception_transformer: nil,
```
jbolinger: I'd like to either keep this in scope for this PR and figure out a better way to do it now, or open an issue so we don't forget to add it back. We do need to support this feature.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
blowmage: This is only a placeholder to demonstrate where the default codes will go. The same for the header params a few lines above. This is only showing where the values will be generated once we can determine what the correct values are.

shared/output/cloud/speech/lib/google/cloud/speech/v1/speech/client.rb
blowmage: Let's track in #107.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_server.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_paged.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
blowmage: This PR is not a WIP. It is not ever intended to be merged. It is for demonstration purposes only.

blowmage:
The real PR for these changes is #108.

Issue #95, opened 2019-04-03
Review Installation Instructions


blowmage:
Review all instructions for getting installed and make sure they still make sense now that we are using the grpc-tools gem for `protoc`

miraleung:
Docs worked for me! Added some Ruby set-up steps in #461 .

PR #96, opened 2019-04-03
Update README.md

quartzmo:
Fix install instructions.

[refs #95]


blowmage:
Does the use of `grpc-tools` remove the need for the "Install the Proto Compiler" section of the README? If we are grabbing `protoc` from the gem we don't also need to have it installed on the system, right?

jbolinger:
@blowmage yes it does remove the need for a system install of protoc.

PR #105, opened 2019-04-12
Run rubocop autocorrect on a dir rather than per file

landrito:
I have been having some pain regenerating the libraries since it takes quite a while since rubocop runs individually per file making it really slow.

#### Before
```sh
$ time bundle exec rake gen
...
> bundle exec rake gen  134.99s user 52.76s system 98% cpu 3:10.42 total
```

#### After
```sh
$ time bundle exec rake gen
...
> bundle exec rake gen  35.73s user 8.23s system 98% cpu 44.503 total
```


blowmage:
I fixed the CI build, but I can't flip CLA back to green. @landrito I'm assigning this to you so you can flip the CLA and merge. Thanks!

PR #106, opened 2019-04-15
Rename Gapic::Template to Gapic::PathTemplate

quartzmo:
To mitigate a major naming conflict with ERB templates:

* Rename `Template` to `PathTemplate`, including variable names, accessors, tests, etc.
* Update documentation with links to URI Template spec.

### DIFF
--- a/gapic-generator/lib/google/gapic/path_template.rb
+++ b/gapic-generator/lib/google/gapic/path_template.rb
@@ -14,18 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "google/gapic/template/parser"
+require "google/gapic/path_template/parser"
 
 module Google
   module Gapic
-    module Template
-      # Parse a URI template.
+    module PathTemplate
+      # Parse a URI path template.
       #
-      # @param template [String] The template to be parsed.
+      # @see https://tools.ietf.org/html/rfc6570 URI Template
       #
-      # @return [Array<Template::Segment|String>] The segments of the template.
-      def self.parse template
-        Parser.new(template).segments
+      # @param path_template [String] The URI path template to be parsed.
+      #
+      # @return [Array<Template::Segment|String>] The segments of the URI path
+      #   template.
+      def self.parse path_template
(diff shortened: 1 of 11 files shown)


gapic-generator/lib/google/gapic/path_template.rb
```diff
@@ -14,18 +14,21 @@
...
       #
-      # @param template [String] The template to be parsed.
+      # @see https://tools.ietf.org/html/rfc6570 URI Template
       #
-      # @return [Array<Template::Segment|String>] The segments of the template.
-      def self.parse template
-        Parser.new(template).segments
+      # @param path_template [String] The URI path template to be parsed.
+      #
+      # @return [Array<Template::Segment|String>] The segments of the URI path
```
blowmage:
```suggestion
      # @return [Array<PathTemplate::Segment|String>] The segments of the URI path
```

gapic-generator/lib/google/gapic/path_template/parser.rb
```diff
@@ -14,57 +14,60 @@
...
+      # @!attribute [r] path_template
+      #   @return [String] The URI path template to be parsed.
       # @!attribute [r] segments
-      #   @return [Array<Segment|String>] The segments of the parsed template.
+      #   @return [Array<Segment|String>] The segments of the parsed URI path
+      #     template.
       class Parser
         # @private
-        # /((?<positional>\*\*?)|{(?<name>[^\/]+?)(?:=(?<template>.+?))?})/
+        # /((?<positional>\*\*?)|{(?<name>[^\/]+?)(?:=(?<path_template>.+?))?})/
```
blowmage: I would keep the regexp the same as it is the same one used in the Python implementation.

gapic-generator/lib/google/gapic/path_template/parser.rb
```diff
@@ -14,57 +14,60 @@
...
         TEMPLATE = %r{
```
blowmage: I would rename the constant instead.

```suggestion
        PATH_TEMPLATE = %r{
```

PR #108, opened 2019-04-17
Update Cloud for Refactor ApiCall

blowmage:
This PR implements the changes in the generator for the GAX changes made in googleapis/gax-ruby#178.

* Refactor how generated client defines methods
* Use ApiCall.new
* Use ApiCall::Options
* Document StreamInput for client streaming requests
* Remove client_config
* Remove exception_transformer

### DIFF
--- a/gapic-generator-cloud/templates/cloud/client/method/_init.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/_init.erb
@@ -1,5 +1,0 @@
-<%- assert_locals method -%>
-<%= method.ivar %> = Google::Gax::ApiCall.new(
-  @<%= method.service.stub_name %>.method(:<%= method.name %>),
-  defaults
-)
--- a/gapic-generator-cloud/templates/cloud/client/method/_init.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/_init.erb
@@ -1,6 +1,5 @@
 <%- assert_locals method -%>
 <%= method.ivar %> = Google::Gax.create_api_call(
   @<%= method.service.stub_name %>.method(:<%= method.name %>),
-  defaults,
-  exception_transformer: exception_transformer
+  defaults
 )
--- a/gapic-generator-cloud/templates/cloud/client/method/_init.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/_init.erb
@@ -1,5 +1,5 @@
 <%- assert_locals method -%>
-<%= method.ivar %> = Google::Gax.create_api_call(
+<%= method.ivar %> = Google::Gax::ApiCall.new(
   @<%= method.service.stub_name %>.method(:<%= method.name %>),
   defaults
 )
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -6,7 +6,7 @@
 <%- end -%>
 # @param requests [Enumerable<<%= method.request_type %> | Hash>]
 #   An enumerable of {<%= method.request_type %>} instances.
-# @param options [Google::Gax::CallOptions]
+# @param options [Google::Gax::ApiCall::Options]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
 # @return [Enumerable<<%= method.return_type %>>]
--- a/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
+++ b/gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
@@ -6,18 +6,22 @@
 <%- end -%>
 # @param requests [Google::Gax::StreamInput, Enumerable<<%= method.request_type %> | Hash>]
 #   An enumerable of {<%= method.request_type %>} instances.
-# @param options [Google::Gax::ApiCall::Options]
+# @param options [Google::Gax::ApiCall::Options, Hash]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
-# @return [Enumerable<<%= method.return_type %>>]
-#   An enumerable of {<%= method.return_type %>} instances.
+# @yield [response] Called on each streaming responses, when provided.
+# @yieldparam response [<%= method.return_type %>]
+#
+# @return [Enumerable<<%= method.return_type %>, Thread>]
+#   An enumerable of {<%= method.return_type %>} instances when a block is not provided.
+#   When a block is provided a thread running the block for every streamed response is returned.
 #
 # @raise [Google::Gax::GaxError] if the RPC is aborted.
 #
 # @example
 <%= indent method.code_example, "#   " %>
 #
-def <%= method.name %> requests, options: nil
+def <%= method.name %> requests, options: nil, &block
   unless requests.is_a? Enumerable
     if requests.respond_to? :to_enum
       requests = requests.to_enum
@@ -30,5 +34,14 @@ def <%= method.name %> requests, options: nil
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
+  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+  retry_policy = {} # retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
+  options.apply_defaults timeout: @timeout, metadata: metadata, retry_policy: retry_policy
+
+  <%= method.ivar %> ||= Google::Gax::ApiCall.new @<%= method.service.stub_name %>.method :<%= method.name %>
+  <%= method.ivar %>.call(requests, options: options, stream_callback: block)
 end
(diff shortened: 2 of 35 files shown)


gapic-generator-cloud/templates/cloud/client/method/_init.erb
```diff
@@ -1,6 +1,5 @@
 <%- assert_locals method -%>
-<%= method.ivar %> = Google::Gax.create_api_call(
+<%= method.ivar %> = Google::Gax::ApiCall.new(
   @<%= method.service.stub_name %>.method(:<%= method.name %>),
```
jbolinger: nit: still worth breaking the line now that there's only a couple params?

shared/Gemfile
```diff
@@ -6,6 +6,6 @@ gem "gapic-generator", path: "../gapic-generator"
 gem "gapic-generator-cloud", path: "../gapic-generator-cloud"
 gem "googleapis-common-protos", [">= 1.3.9", "< 2.0"]
 gem "googleapis-common-protos-types", [">= 1.0.4", "< 2.0"]
-gem "google-gax", "~> 1.0"
+gem "google-gax", github: "blowmage/gax-ruby", branch: "2.0/refactor-api_call"
```
blowmage: Before this is merged I would like to change this to reference master. So the GAX PR should be merged first, and we should see one green build referencing master before merging.

gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
```diff
@@ -4,27 +4,44 @@
...
   end
 
   requests = requests.lazy.map do |request|
     Google::Gax.to_proto request, <%= method.request_type %>
   end
 
-  <%= method.ivar %>.call(requests, options)
+  # Converts hash and nil to an options object
+  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+  header_params = {} # { name: request.name }
```
jbolinger: Are you planning to implement this part later?

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
```diff
@@ -4,13 +4,13 @@
...
-# @param requests [Enumerable<<%= method.request_type %> | Hash>]
+# @param requests [Google::Gax::StreamInput, Enumerable<<%= method.request_type %> | Hash>]
 #   An enumerable of {<%= method.request_type %>} instances.
-# @param options [Google::Gax::CallOptions]
+# @param options [Google::Gax::ApiCall::Options, Hash]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc.
 #
-# @yield [result, operation] Access the result along with the RPC operation
-# @yieldparam result [<%= method.return_type %>]
+# @yield [response, operation] Access the result along with the RPC operation
```
jbolinger: nit: use response instead of result in the text for consistency? Should we note the concrete type of the `operation`?

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
```diff
@@ -29,15 +29,16 @@
 #     named arguments.
 <%- end -%>
 #
-# @yield [operation] Register a callback to be run when an operation is done.
-# @yieldparam operation [Google::Gax::Operation]
+# @yield [response, operation] Access the result along with the RPC operation
```
jbolinger: nit: same as before, but there is where I'm worried folks might get confused (lro operation vs grpc operation) and why I think linking to the gRPC type might help clarify.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
```diff
@@ -46,13 +47,19 @@ def <%= method.name %> request = nil, options: nil, **request_fields
   end
 
   request ||= request_fields
+  # request = Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
```
jbolinger: ?

gapic-generator-cloud/templates/cloud/client/method/def/_normal.erb
```diff
@@ -47,7 +47,19 @@ def <%= method.name %> request = nil, options: nil, **request_fields, &block
   end
 
   request ||= request_fields
+  # request = Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
```
jbolinger: ?

gapic-generator-cloud/templates/cloud/client/method/def/_server.erb
```diff
@@ -46,7 +50,17 @@ def <%= method.name %> request = nil, options: nil, **request_fields
   end
 
   request ||= request_fields
+  # request = Google::Gax::Protobuf.coerce request, to: <%= method.request_type %>
```
jbolinger: ?

gapic-generator-cloud/templates/cloud/client/method/def/_bidi.erb
blowmage: Yes, @quartzmo is working on a PR now to move this functionality into the generator.

gapic-generator-cloud/templates/cloud/client/method/def/_client.erb
blowmage: For consistency, yes. Now that we have a `request` argument, it seems more correct to have a `response` returned.

blowmage: > Should we note the concrete type of the `operation`?

The type is documented 2 lines below.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: Right. We discussed this as a breaking change. The new client method's block would be used to peek at the GRPC ActiveCall operation, and not to define the `Gax::Operation#on_done` callback. The type is defined 2 lines below.

Perhaps it would be less confusing if we prefixed `grpc` onto this argument name? So use `grpc_operation` instead of `operation`?

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: This is a future change, I thought I had removed all of them. Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_normal.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_server.erb
blowmage: Removed.

gapic-generator-cloud/templates/cloud/client/method/_init.erb
blowmage: Lol! this template is no longer being used, and should have been deleted. Removed.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: Or, is there a specific value we can pass tot he proc instead of the entire GRPC ActiveCall Operation object?

jbolinger: I can't think of anything better to use. The only real use case we have now is for headers (I think), but I'm not sure breaking it apart would help.

blowmage: Right, so if we only sent the headers this could be renamed `trailing_headers_callback` and send `response` and `headers` arguments.

blowmage:
The Gemfile in shared has been updated to use gax-ruby's master branch. PTAL!

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
jbolinger: I don't have a strong preference either way. I think returning the operation is okay if we document it carefully. The metadata idea is fine too, but if there are other properties that we want to expose in the future we'll have to keep expanding the argument list.

shared/output/gapic/templates/vision/google/cloud/vision/v1/image_annotator.rb
```diff
@@ -119,12 +119,12 @@ def initialize \
 
             defaults = default_settings client_config, timeout, metadata, lib_name, lib_version
 
-            @batch_annotate_images = Google::Gax.create_api_call(
+            @batch_annotate_images = Google::Gax::ApiCall.new(
```
jbolinger: This is still here because only the cloud templates were updated - correct (and same for other test clients)?

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: GRPC might have a wrapper object that restricts what users can change on the object, that they pass to interceptors. Would it be a good idea to use that here?

jbolinger: I'm not sure, do you have an example?

shared/output/gapic/templates/vision/google/cloud/vision/v1/image_annotator.rb
blowmage: Correct. This file shouldn't have been changed, but its going away soon so I'm not too worried about it.

gapic-generator-cloud/templates/cloud/client/method/def/_lro.erb
blowmage: Hah! It looks like Operation is already the "view" object for ActiveCall. This is the right thing to use.

PR #115, opened 2019-04-22
Method presenter routing

### LINKED ISSUE #104: Set params_extractor for Google::Gax.create_api_call
In order for gRPC API backends to route traffic efficiently, it is necessary for some values to be sent in HTTP headers as well as in the request payload.

Code generators should look at URL-based variables declared in the `google.api.http` annotation and transcribe these into the `x-goog-request-params` header in unary calls. A URL-based variable is a variable declared as a key in curly braces in the URI string. For example:

```
rpc CreateTopic(CreateTopicRequest) {
  option (google.api.http).post = "{parent=projects/*}/topics";
}
```

In this case, the applicable variable is parent, and it refers to the parent field in CreateTopicRequest.

Example from google-cloud-spanner:

```rb
                @list_databases = Google::Gax.create_api_call(
                  @database_admin_stub.method(:list_databases),
                  defaults["list_databases"],
                  exception_transformer: exception_transformer,
                  params_extractor: proc do |request|
                    {'parent' => request.parent}
                  end
                )
```

quartzmo:
* Add `MethodPresenter#routing_params` 
  * Return segment names from `google.api.http.*` annotation.
  * Update method def templates
* Add `_options_defaults.erb` partial with `MethodPresenter#routing_params`.
  * Updated expected output for real `google.api.http` annotations.

[closes #104]


blowmage:
@quartzmo You changed the gapic generator expected output, but you didn't change the cloud generator's default output. Since cloud is inheriting the gapic generator's templates, this changes the cloud output.

gapic-generator/templates/default/client/method/def/_options_defaults.erb
```diff
@@ -0,0 +1,18 @@
...
+<%- method.routing_params.each do |rp| -%>
+  "<%= rp %>" => request.<%= rp %>,
+<%- end -%>
+}
+request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
+<%- else -%>
+metadata = @metadata.dup
+<%- end -%>
+retry_policy = {} # retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: Why is this comment here?

gapic-generator/templates/default/client/method/def/_normal.erb
```diff
@@ -49,15 +49,7 @@ def <%= method.name %> request = nil, options: nil, **request_fields, &block
...
   request = Google::Gax.to_proto request, <%= method.request_type %>
 
-  # Converts hash and nil to an options object
-  options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
-
-  # Customize the options with defaults
-  header_params = {} # { name: request.name }
-  request_params_header = header_params.map { |k, v| "#{k}=#{v}" }.join("&")
-  metadata = @metadata.merge "x-goog-request-params".freeze => request_params_header
-  retry_policy = {} # retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
quartzmo: @jbolinger I moved the line with the comment as-is from the `method/def` templates. @blowmage can we remove this comment?

blowmage: That will get removed when we implement retry codes. Let's leave it there for now.

PR #118, opened 2019-04-24
Add Method retry codes

### LINKED ISSUE #117: Configure RPC calls with the correct retry codes

blowmage:
* Add MethodPresenter#retry_codes
  * Return strings of GRPC::Core::StatusCodes values.

[closes #117]

### DIFF
--- a/gapic-generator/templates/default/helpers/presenters/method_presenter.rb
+++ b/gapic-generator/templates/default/helpers/presenters/method_presenter.rb
@@ -141,6 +141,32 @@ class MethodPresenter
     routing_params.any?
   end
 
+  ##
+  # Returns the `GRPC::Core::StatusCodes` that should be retried for this method.
+  #
+  # Strings are returned instead of the actual `GRPC::Core::StatusCodes` values because those constants point to
+  # integers, and then an array of integers would be generated. Returning strings here allows us to generate arrays of
+  # the actual constants which should read better.
+  #
+  # @return [Array<String>] The retry codes
+  #
+  def retry_codes
+    if @method.http && !@method.http.get.empty?
+      # Any RPC with a google.api.http annotation with a non-empty get key (ignore additional_bindings for this purpose)
+      # is assumed to be idempotent and should additionally retry ABORTED, and UNKNOWN errors.
+      return ["GRPC::Core::StatusCodes::UNAVAILABLE",
+              "GRPC::Core::StatusCodes::ABORTED",
+              "GRPC::Core::StatusCodes::UNKNOWN"]
+    end
+
+    # Retry UNAVAILABLE on all methods.
+    ["GRPC::Core::StatusCodes::UNAVAILABLE"]
+  end
+
+  def retry_codes?
+    retry_codes.any?
+  end
+
   protected
 
   def message_ruby_type message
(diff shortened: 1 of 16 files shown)


gapic-generator/templates/default/helpers/presenters/method_presenter.rb
```diff
@@ -141,6 +141,32 @@ def routing_params?
...
+  ##
+  # Returns the `GRPC::Core::StatusCodes` that should be retried for this method.
+  #
+  # Strings are returned instead of the actual `GRPC::Core::StatusCodes` values because those constants point to
+  # integers, and then an array of integers would be generated. Returning strings here allows us to generate arrays of
+  # the actual constants which should read better.
+  #
+  # @return [Array<String>] The retry codes
+  #
+  def retry_codes
```
jbolinger: Can we move this implementation to the cloud generator? I'll defer to @dazuma if these are sensible defaults in that context (`UNAVAILABLE` likely needs to be removed).

For the base generator, I think it would be better to disable retries and return an empty list here.

blowmage: I have removed UNAVAILABLE.

Why do we want to retry in cloud but not retry in default?

jbolinger: Basically because we can't agree on a sensible default retry strategy in the general case yet. For Cloud APIs, we may be able to make some assumptions.  Of course, once the spec is updated we can revisit this.

jbolinger:
@blowmage I'd also be okay with leaving this as-is in the base generator and just keeping #117 open nothing that we're blocked on a decision for what the default policy should be. WDYT?

blowmage:
I'd prefer we implemented the previous design with the acknowledgment that is is likely to change. But I would prefer to have some implementation, rather than leaving it unimplemented, or guessing at what a final design will be.

jbolinger:
@blowmage I'm not sure what you mean by previous design. I'm fine approving this PR as-is (i.e. without `UNAVAILABLE`) under the assumption that we'll keep the issue open to track the spec changes. It's still a best guess, but that's the current consensus of what the final decision will be. I'm not sure how we can do any better than that?

IMHO retrying by default for any arbitrary API is more confusing than it's worth and I'd still want to move this to the Cloud generator. But, I don't really care that much as long as we document it since this is easy to customize. So, that bit I'll leave that up to you. Retrying on `UNAVAILABLE`, in either case, doesn't make much sense to me though. It's not likely to be correct in the future and it's wrong now.

blowmage:
> I'm not sure what you mean by previous design.

By "previous design" I mean the behavior currently described in the design document. I get that things are in flux and subject to change, but I'm not privy to those discussions, and so far the design document hasn't been updated. I would rather implement what is in the design document, and change the implementation when the design document changes.

blowmage:
>  I'd still want to move this to the Cloud generator

I don't understand why the cloud generator should diverge from default generator's retry behavior. What is the use case for this? Is there something different about the GCP APIs that warrant different retry semantics?

jbolinger:
> I would rather implement what is in the design document, and change the implementation when the design document changes.

I get it, but we have to live with out of date docs. You can ping the owner and try to get it updated 😄. I'd still much rather implement what is most likely to be correct now. We'll have to track it via an issue either way until it's decided so why do something that we know won't work *and* is not likely to be correct long term?

> Is there something different about the GCP APIs that warrant different retry semantics?

Yeah, we assume that they follow various patterns and go through a review process. In this case, we're assuming folks writing a service that's exposed via a http `GET` method will be idempotent. Of course, they generally should be, but in practice a user can do anything that they want.

PR #125, opened 2019-04-26
Move Protobuf Docs to proto_docs directory

### LINKED ISSUE #123: Documentation directory outside of lib?
The generator currently creates a bunch of documentation files in the `/lib/doc` directory. This location should be fine, as we are not expecting anyone to require those file paths. However, it is still possible. Currently we have added a `raise` statement with an error message stating that these files are not intended to be loaded. That said, what about placing this directory outside of `/lib`, and configuring the `.yardoc` file to load the files in that directory as well? That would allow YARD to know about the documentation, but avoid the issues of accidentally requiring the files. Thoughts?

blowmage:
This PR moved the generated documentation files outside of `/lib` so it can't be required or loaded.

[closes #123]


quartzmo:
Any other suggestions for the name of the new directory?

blowmage:
I have a preference for simple names, but I didn't want to use `/doc` or `/docs` because I didn't want to conflict with YARD or RDoc. So I ended up with `/proto_docs` which felt the most descriptive. Do you have any other suggestions?

quartzmo:
I really tried, but I can't think of anything great. Anyone else?

PR #126, opened 2019-04-30
SCOPE Refactor

### LINKED ISSUE #102: Move ALL_SCOPES
Right now this is a constant on the Client class. It may make more sense for these values to be on the Credentials class, since Credentials is now service-specific and that is where they are used.

blowmage:
Since Credentials is now part of a service's generated output,
use Credentials::SCOPE and remove Client::ALL_SCOPES.
it makes more sense for the credentials to manage these values.

[closes #102]


gapic-generator-cloud/templates/cloud/service/client/_credentials.erb
```diff
@@ -3,7 +3,7 @@
 require "googleauth"
 <% end %>
 class <%= service.credentials_name %> < Google::Auth::Credentials
-  SCOPE = ["https://www.googleapis.com/auth/cloud-platform"]
+  SCOPE = <%= service.client_scopes.inspect %>
```
quartzmo: Can we freeze the assigned value of `SCOPE` here (and similarly in `gapic-generator`)?

blowmage: The Rubocop formatting will add it. This will go away eventually, so I don't really care to add a `.freeze` here.

quartzmo: Sounds good. Want to add a comment explaining where this is headed/ why it will go away?

PR #143, opened 2019-05-15
Allow accessing trailing metadata on streaming responses

blowmage:
This PR allows for accessing trailing metadata on streaming responses. It does this by removing the `stream_callback` callback and normalizes all return values and closure arguments to be the same, even for streaming responses.

For example, let's assume a client has an RPC method `streaming_rpc` that returns a stream of responses. Before this change, users could use the method's closure to process each streamed response:

```ruby
client.streaming_rpc stream_input do |response|
  process_response response
end
```

With this PR users will not need to iterate over the Enumerable that is returned from the RPC method:

```ruby
responses = client.streaming_rpc stream_input

Thread.new do
  responses.each do |response|
    process_response response
  end
end
```

Why does this have to change? Because the only way we to expose trailing metadata is by using the method closure to pass both the response and the gRPC Operation object that contains the trailing metadata:

```ruby
client.streaming_rpc stream_input do |responses, operation|
  # pull off trailing metadata from the grpc operation
end
```

Therefore, all client RPC methods now use the same API for RPC method closures. Whether it is a normal response, streaming response, paginated response, LRO Operation, etc...


blowmage:
This PR has been rebased on latest master.

quartzmo:
@blowmage 

1. What are the pros and cons of this approach?

2. It might help to add before and after examples to the PR, since it's pretty hard to quickly grasp the change given the number of files in this PR.

blowmage:
PR description has been updated with example code demonstrating the changes.

PR #146, opened 2019-05-20
Fix Showcase acceptance tests

blowmage:
The tests use GRPC, but Gax recently fixed loading grpc too early.
Because of this, the tests need to require grpc in order to use it.


blowmage:
These changes were included in #145, which was accepted and merged. Therefore this PR can be closed.

PR #155, opened 2019-05-30
Switch Client to positional args

blowmage:
This PR changes the Client RPC methods to use positional arguments instead of the mixed positional/named arguments. This is needed to work around Protobuf [adding the `#to_hash` method to the resource classes](https://github.com/protocolbuffers/protobuf/pull/6166).

This PR is dependent on googleapis/gax-ruby#208.

### DIFF
--- a/gapic-generator/templates/default/service/client/method/docs/_request_normal.erb
+++ b/gapic-generator/templates/default/service/client/method/docs/_request_normal.erb
@@ -3,5 +3,12 @@
 <%- if method.doc_description -%>
 <%= indent method.doc_description, "#   " %>
 <%- end -%>
-# @param options [Google::Gax::ApiCall::Options, Hash]
+#
+#   When using a hash, the following fields are supported:
+#
+<%- method.arguments.each do |arg| -%>
+#   * `<%= arg.name %>` (`<%= arg.doc_types %>`)<%- if arg.doc_description -%>:
+<%= indent arg.doc_description, "#     " %>
+<%- end -%>
+<%- end -%># @param options [Google::Gax::ApiCall::Options, Hash]
 #   Overrides the default settings for this call, e.g, timeout, retries, etc. Optional.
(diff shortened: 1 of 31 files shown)


gapic-generator/templates/default/service/client/method/docs/_request_normal.erb
```diff
@@ -1,25 +1,7 @@
...
-#   @param options [Google::Gax::ApiCall::Options, Hash]
-#     Overrides the default settings for this call, e.g, timeout, retries, etc.
-<%- if method.arguments.find { |arg| arg.name == "options" } -%>
-#
-#     If you want to provide a {<%= method.request_type %>#options}
-#     value you will need to pass all attributes in a Hash instead of using
-#     named arguments.
+<%= indent method.doc_description, "#   " %>
 <%- end -%>
+# @param options [Google::Gax::ApiCall::Options, Hash]
```
quartzmo: Add `nil` to the`options` types list?

blowmage: Since `nil` is the default value, we usually leave it off. Although we add "Optional."

blowmage: Added "Optional." to the `options` documentation.

gapic-generator/templates/default/service/client/method/docs/_request_normal.erb
```diff
@@ -3,5 +3,12 @@
 <%- if method.doc_description -%>
 <%= indent method.doc_description, "#   " %>
 <%- end -%>
-# @param options [Google::Gax::ApiCall::Options, Hash]
+#
+#   When using a hash, the following fields are supported:
+#
+<%- method.arguments.each do |arg| -%>
+#   * `<%= arg.name %>` (`<%= arg.doc_types %>`)<%- if arg.doc_description -%>:
```
quartzmo: What do you think about using square brackets for the types list, like YARD does?

blowmage: This is markdown content, and not using YARD markup. Users won't ever see the square brackets used by YARD. I thought it would make more sense to see the type in parens on the generated documentation.

PR #157, opened 2019-05-30
 Use Gax header helper

blowmage:
Move this common code to Gax so the GAPIC client can do less. This is dependent on googleapis/gax-ruby#210.


blowmage:
> Without running tests for the clients, it's hard for me to be certain that this change will work as intended in the generated clients

CI runs the showcase acceptance tests, which generate a client and use it to make calls. So if we have a green build I'm fairly confident in it working.

PR #180, opened 2019-06-27
Add gapic-common gem

blowmage:
Replaces google-gax 2.0 in order to avoid dependency conflicts.


blowmage:
@dazuma Should we remove the coverage stuff before merging? Or would you like to have this on all PRs moving forward?

PR #197, opened 2019-08-09
Add gen task to gapic gem

blowmage:


blowmage:
The top-level `rake gen` task breaks because this task is missing. We have similar no-op tasks in gapic-common and shared.

PR #198, opened 2019-08-09
Rename CallOptions

blowmage:
* Move CallOptions up a namesapce
* Nest RetryPolicy under CallOptions
* Move ApiCall under Grpc::Stub


jbolinger:
Looks like the showcase tests need to be updated as well?

blowmage:
> don't you ever get tired of renaming things 😄?

I do not.

PR #200, opened 2019-08-09
Add showcase Collect test

jbolinger:
Add the final RPC, `Collect`, from the showcase `Echo` service to the tests.

~(Also remove a few prints that probably made it in by accident).~


shared/test/showcase/echo/chat_test.rb
```diff
@@ -32,7 +32,6 @@ def test_chat
 
     Thread.new do
       10.times do |n|
-        puts "PUSH hello #{n}:1"
```
blowmage: Can we add a conditional instead of removing these? I prefer to see them.

```ruby
puts "PUSH hello #{n}:1" if ENV["VERBOSE"]
```

jbolinger: I don't mind leaving them if you actually want them. I assumed it was just an accident - they drive me crazy though 😄. I'll revert it.

PR #208, opened 2019-08-13
Add gem dashed and slashed files

blowmage:
Cloud defines these dashed and slashed files. As we work towards supporting more current Cloud-specific features we will want these files.

This PR also defines a Gem-level configuration that can set common values above the package/version level or the service/client level.

### DIFF
--- a/gapic-generator-cloud/templates/cloud/service/client/_client.erb
+++ b/gapic-generator-cloud/templates/cloud/service/client/_client.erb
@@ -1,0 +1,110 @@
+<%- assert_locals service -%>
+<% @requires = capture do %>
+require "gapic/common"
+require "gapic/config"
+require "gapic/config/method"
+
+require "<%= service.gem.name.gsub "-", "/" %>"
+require "<%= service.gem.version_require %>"
+require "<%= service.proto_service_require %>"
+require "<%= service.credentials_require %>"
+<%- if service.paths? -%>
+require "<%= service.paths_require %>"
+<%- end -%>
+<%- if service.lro? -%>
+require "<%= service.operations_require %>"
+<%- end -%>
+<% end %>
+# Service that implements <%= service.name %> API.
+class <%= service.client_name %>
+<%- if service.paths? -%>
+  include <%= service.paths_name %>
+
+<%- end -%>
+  # @private
+  attr_reader :<%= service.stub_name %>
+
+  ##
+  # Configuration for the <%= service.name %> <%= service.client_name %> API.
+  #
+  # @yield [config] Configure the <%= service.client_name %> client.
+  # @yieldparam config [<%= service.client_name %>::Configuration]
+  #
+  # @return [<%= service.client_name %>::Configuration]
+  #
+  def self.configure
+    @configure ||= <%= service.client_name %>::Configuration.new <%= service.gem.namespace %>.configure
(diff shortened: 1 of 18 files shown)


gapic-generator-cloud/templates/cloud/service/client/_client.erb
```diff
@@ -0,0 +1,110 @@
+<%- assert_locals service -%>
+<% @requires = capture do %>
+require "gapic/common"
+require "gapic/config"
+require "gapic/config/method"
+
+require "<%= service.gem.name.gsub "-", "/" %>"
```
jbolinger: There is very little difference in this file from the base file. Do you think it's possible to avoid overriding the whole file for these customizations?

shared/output/cloud/showcase/lib/google-showcase.rb
```diff
@@ -0,0 +1 @@
+require "google/showcase"
```
jbolinger: Do you think we should do showcase with the cloud or base generator? I think it may be better to use the base templates for showcase and real APIs for the cloud generator. WDYT?

(Whatever we decide let's leave this the way it is for this PR)

gapic-generator-cloud/templates/cloud/service/client/_client.erb
blowmage: Possible, but I don't know what those changes should be just yet. I'd like to get further along getting a real, working cloud gem with all the expected dependencies and then figure out how to avoid duplication.

shared/output/cloud/showcase/lib/google-showcase.rb
blowmage: The acceptance tests run off of the default generator. The fact that we are still testing showcase on the cloud generator is an artifact to back when cloud was the only working generator and default was using a different implementation. Pre #109.

gapic-generator-cloud/templates/cloud/service/client/_client.erb
jbolinger: Ok, that's fair. Let's not let this get out of hand but we can live with one or two of these kinds of things for awhile.

shared/output/cloud/showcase/lib/google-showcase.rb
jbolinger: Hm, I'm still confused. If it's running off of the default generator then this extra file shouldn't exist? I see the `Rakefile` has a TODO for the template set so is something still hardwired or why does this file get generated?

blowmage: The acceptance tests run the generator when the tests are run, and don't use these files. These files are used to test the cloud generator output to ensure it is working as expected.

There are features that right now only showcase uses, and in order to get the broadest test coverage we were testing the cloud generator against the showcase protobufs.

jbolinger: Ah, I see - thanks for clarifying.

blowmage:
> Do we want to add the `endpoint` param?

We do. We already have it on the Client/Service configuration. But a gem can contain multiple Clients/Services, and they can all have different endpoints. So my thinking is that we shouldn't be providing Client/Service specific-configuration at the gem-level configuration.

PR #212, opened 2019-08-15
Update submodules to recent

blowmage:
Get the latest versions of the protobuf files and services. Update to use the latest protobuf annotation changes.


gapic-generator/lib/gapic/schema/wrappers.rb
```diff
@@ -379,7 +371,7 @@ def ruby_package
       # @return [Google::Api::Package] Packaging information.
       #   See `google/api/client.proto`.
       def client_package
-        options[:".google.api.client_package"] if options
+        options[:".google.api.package"] if options
```
jbolinger: Does this actually exist? It's no longer in `google/api/client.proto` as the line above states.

.gitmodules
```diff
@@ -1,9 +1,9 @@
...
-	url = https://github.com/protocolbuffers/protobuf.git
-[submodule "protos/googleapis"]
+	url = git@github.com:googleapis/api-common-protos.git
+[submodule "shared/googleapis"]
 	path = shared/googleapis
-	url = https://github.com/googleapis/googleapis.git
+	url = git@github.com:googleapis/googleapis.git
+[submodule "shared/protobuf"]
+	path = shared/protobuf
+	url = git@github.com:protocolbuffers/protobuf.git
```
jbolinger: Do we prefer to use git urls vs https?

gapic-generator/lib/gapic/schema/wrappers.rb
blowmage: This method was supposed to be removed but wasn't. Removed.

.gitmodules
blowmage: I had a hard time updating the submodules cleanly, so i removed and re-added them.

blowmage: I didn't change this on purpose, that was just how i added them back. I think the git protocol is considered faster than https, so I just always default to using it.

jbolinger: 👍 that works for me.

gapic-generator/templates/default/helpers/presenters/service_presenter.rb
```diff
@@ -176,10 +176,11 @@ def helpers_require
 
   def references
     @references ||= begin
-      m = @service.parent.messages.select { |m| m.fields.select(&:resource).any? }
-      pairs = m.map { |m1| [m1.name, m1.fields.map(&:resource).compact.first.pattern] }
-      pairs.sort_by! { |name, tmplt| name }
-      pairs.map { |name, tmplt| ResourcePresenter.new name, tmplt }
+      m = @service.parent.messages.select(&:resource)
+      m.sort_by! { |m1| m1.name }
+      # TODO: ResourceDescriptor#pattern is a list of patterns
```
jbolinger: This is fine. Let's defer it. 

I think this [history bit](https://github.com/googleapis/api-common-protos/blob/2769b82d4993ccbe66a3d886aa5466fdbd050ea2/google/api/resource.proto#L104) is new too and I find it confusing.

blowmage: If/when this PS is merged I will open an issue to track this.

PR #215, opened 2019-08-20
Add tests for showcase operations client

jbolinger:
Ads tests for new methods in showcase 0.4.0 and consolidate all the `Google::Showcase::V1beta1::Echo::Client.new` calls.

Note that `list_operations` produces a oneof deserialization error when used with the `page_size` parameter. I don't think it's critical that these tests cover that (it may be a server side issue), but I could not figure out why the error is thrown.

### DIFF
--- a/shared/test/showcase/echo/operations_test.rb
+++ b/shared/test/showcase/echo/operations_test.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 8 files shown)


shared/test/showcase/echo/operations_test.rb
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+require "base64"
+require "test_helper"
+require "google/showcase/v1beta1/echo"
+require "grpc"
+
+class EchoTest < ShowcaseTest
+  def test_raise_invalid_get_operation
+    client = new_operations_client
```
quartzmo: You can refactor all of the `client = new_operations_client` statements to a single `before` method if you want.

jbolinger: Good idea - how's this?

shared/test/test_helper.rb
```diff
@@ -39,6 +39,18 @@ def generate_library_for_test imports, protos
 end
 
 class ShowcaseTest < Minitest::Test
+  def new_client
```
blowmage: Would the name `echo_client` be more precise? Especially if we add tests for the non-Echo services in Showcase?

jbolinger: Yeah, I suppose so. You mean helpers like `new_echo_client`, `new_echo_operations_client`, `new_identify_client`, etc.?

blowmage: Yep.

jbolinger: Sure, I'll send a follow up shortly.

PR #228, opened 2019-08-30
Expose operation to paged RPC calls on all pages

### LINKED ISSUE #227: gRPC operation should be accessible in all pages of a PagedEnumerable
It is currently only available on the first page of results via the operation block.

jbolinger:
The underlying gRPC operation should always be available so that the `trailing_metadata` can be accessed by clients. Currently, this is only possible on the first page of a paginated RPC and this change adds an `operation` property to the `Page` class and exposes the `operation` via the `format_response` callback.

Note that the `format_response` callback returns a nil operation for streaming methods - I'm not sure if we want to also pass the operation to this callback each time or document it as-is. Thoughts?

[resolves #227]

### DIFF
--- a/gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
+++ b/gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
@@ -135,7 +135,7 @@ module Gapic
             Thread.new { operation.execute }
           else
             response = operation.execute
-            response = format_response.call response if format_response
+            response = format_response.call response, operation if format_response
             operation_callback&.call response, operation
             response
           end
@@ -156,7 +156,7 @@ module Gapic
         return unless stream_callback
         return stream_callback unless format_response
 
-        proc { |response| stream_callback.call format_response.call response }
+        proc { |response| stream_callback.call format_response.call(response, nil) }
       end
 
       def calculate_deadline options
--- a/gapic-common/lib/gapic/paged_enumerable.rb
+++ b/gapic-common/lib/gapic/paged_enumerable.rb
@@ -52,11 +52,12 @@ module Gapic
     # @param method_name [Symbol] The RPC method name.
     # @param request [Object] The request object.
     # @param response [Object] The response object.
+    # @param operation [GRPC::ActiveCall::Operation] the current RPC operation.
     # @param options [Gapic::CallOptions] The options for making the RPC call.
     # @param format_resource [Proc] A Proc object to format the resource object. The Proc should accept response as an
     #   argument, and return a formatted resource object. Optional.
     #
-    def initialize grpc_stub, method_name, request, response, options, format_resource: nil
+    def initialize grpc_stub, method_name, request, response, operation, options, format_resource: nil
       @grpc_stub = grpc_stub
       @method_name = method_name
       @request = request
@@ -68,7 +69,7 @@ module Gapic
       verify_request!
       verify_response!
 
-      @page = Page.new @response, @resource_field
+      @page = Page.new @response, @resource_field, operation
     end
 
     ##
@@ -83,7 +84,7 @@ module Gapic
 
       each_page do |page|
         page.each do |obj|
-          obj = @format_resource.call obj if @format_resource
+          obj = @format_resource.call obj, page.operation if @format_resource
           yield obj
         end
       end
@@ -124,9 +125,10 @@ module Gapic
 
       next_request = @request.dup
       next_request.page_token = @page.next_page_token
-      next_response = @grpc_stub.call_rpc @method_name, next_request, options: @options
 
-      @page = Page.new next_response, @resource_field
+      operation_callback = ->(next_response, operation) { @page = Page.new next_response, @resource_field, operation }
+      @grpc_stub.call_rpc @method_name, next_request, options: @options, operation_callback: operation_callback
+      @page
     end
 
     ##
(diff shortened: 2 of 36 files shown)


gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
```diff
@@ -135,7 +135,7 @@ def call request, options: nil, format_response: nil, operation_callback: nil, s
             Thread.new { operation.execute }
           else
             response = operation.execute
-            response = format_response.call response if format_response
+            response = format_response.call response, operation if format_response
```
blowmage: I don't understand why `operation` is being added to the `format_response` callback when the `operation_callback` already exists. Shouldn't `operation_callback` be called for every paged request/response instead?

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -52,11 +52,12 @@ class PagedEnumerable
     # @param method_name [Symbol] The RPC method name.
     # @param request [Object] The request object.
     # @param response [Object] The response object.
+    # @param operation [GRPC::ActiveCall::Operation] the current RPC operation.
```
blowmage: I think adding `operation` to this object is good. I think we could add additional `each_with_operation` and `each_page_with_operation` methods to make it easier for users to access this information.

gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
jbolinger: This is because `format_response` is used to create the `PagedEnumerable`. See `gapic-generator/templates/default/service/client/method/def/_response.erb`.

It is a bit odd to include here, but I didn't see an obvious alternative and I don't think it hurts.

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -52,11 +52,12 @@ class PagedEnumerable
...
     # @param options [Gapic::CallOptions] The options for making the RPC call.
     # @param format_resource [Proc] A Proc object to format the resource object. The Proc should accept response as an
     #   argument, and return a formatted resource object. Optional.
     #
-    def initialize grpc_stub, method_name, request, response, options, format_resource: nil
+    def initialize grpc_stub, method_name, request, response, operation, options, format_resource: nil
```
blowmage:
```suggestion
    def initialize grpc_stub, method_name, request, response, operation, options, format_resource: nil, operation_callback: nil
```

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -83,7 +84,7 @@ def each
 
       each_page do |page|
         page.each do |obj|
-          obj = @format_resource.call obj if @format_resource
+          obj = @format_resource.call obj, page.operation if @format_resource
```
blowmage:
```suggestion
          obj = @format_resource.call obj if @format_resource
```

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -124,9 +125,10 @@ def next_page
 
       next_request = @request.dup
       next_request.page_token = @page.next_page_token
-      next_response = @grpc_stub.call_rpc @method_name, next_request, options: @options
```
blowmage: I would create a new operation callback that invokes the user callback if provided, and captures the operation object to it can be stored on the Page:

```ruby
next_operation = nil
next_operation_callback = ->(next_resp, next_op) do
  # call the user-supplied callback if present.
  operation_callback.call next_resp, next_op if @operation_callback
  # save the operation to next_operation so we can use it.
  next_operation = next_op
end
next_response = @grpc_stub.call_rpc @method_name, next_request, options: @options
next_response = @grpc_stub.call_rpc @method_name, next_request, options: @options, operation_callback: next_operation_callback

@page = Page.new next_response, next_operation, @resource_field
```

gapic-common/lib/gapic/paged_enumerable.rb
blowmage: I kinda think the `@format_response.call` should be pushed down to `Page#each`.

gapic-common/lib/gapic/paged_enumerable.rb
jbolinger: I added a `each_page_with_operation` but I'm not sure a `each_with_operation` really makes sense?

gapic-common/lib/gapic/paged_enumerable.rb
jbolinger: done

gapic-common/lib/gapic/paged_enumerable.rb
jbolinger: ok, moved.

gapic-common/lib/gapic/paged_enumerable.rb
jbolinger: how's this?

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -107,6 +109,25 @@ def each_page
...
     end
 
+    ##
+    # Iterate over the pages with the operation used to fetch the page.
+    #
+    # @yield [Page] Gives the pages in the stream.
+    #
+    # @raise [GapicError] if it's not started yet.
+    #
+    def each_page_with_operation
```
blowmage: Since `#operation` is already on `Page`, I don't think this offers much utility to the user.

My thought was to have a resource-centered method that also included operation. Sort of like [`Enumerable#each_with_index`](https://ruby-doc.org/core-2.6.4/Enumerable.html#method-i-each_with_index). But, the more I have thought about this, the more I wonder how it should behave. Does it return the same `operation` for each resource in the page? Or does it return the `operation` for only the first resource on the page?  What if the page has an empty list of resources but has `operation`? (Is that even possible?)

I'd be fine removing `each_page_with_operation`. I think I would still favor a `each_with_operation` method, but we should probably discuss the behavior in more detail.

blowmage:
This PR's functionality has been implemented in #231, #232, and #234. Closing

PR #229, opened 2019-08-30
Add support for multiple samplegen files

blowmage:
This varies slightly from the spec, becuase of difficulties getting protoc to
recognize multiple files.
The samplegen response now allows for more than just loop.


gapic-generator/templates/default/helpers/presenters/sample_presenter.rb
```diff
@@ -45,16 +45,8 @@ def kwargs
...
-    @sample_config["response"][0]["loop"]["variable"]
-  end
-
-  def response_collection
-    @sample_config["response"][0]["loop"]["collection"].gsub "$resp", "response"
-  end
-
-  def response_body
-    @sample_config["response"][0]["loop"]["body"]
+  def response_raw
```
blowmage: This is greatly simplified because these methods were specific to the original samplegen file, and didn't work on the second file. I didn't do any wrapping of the values because I'm still not sure how this will end up. This is very much still a WIP.

gapic-generator/templates/default/service/client/method/docs/sample_response/_loop.erb
```diff
@@ -0,0 +1,14 @@
+<%- assert_locals values -%>
+<%= values["collection"].gsub("$resp", "response") %>.each do |result|
+  <%- values["body"].each do |b| -%>
+  <%- if b.key? "comment" -%>
```
blowmage: We could probably reuse the `service/client/method/docs/sample_response/_comment.erb` partial here, but I left this code alone for now. It is not much changed from the code extracted from `service/client/method/docs/sample.erb`.

gapic-generator/templates/default/helpers/presenters/sample_presenter.rb
jbolinger: What is the original samplegen file? Are the two that we're working off of now consistent (i.e. the files with `schema_version: 1.2.0`)?

blowmage: Yes, the files we have are 1.2.0. I made this change to support the second samplegen file you sent us (added in this PR). The presenter methods we had before couldn't work with the second file, so I named this method with `raw` to indicate that the abstraction is incomplete.

jbolinger: Ok, I think it's fine to leave this as-is for awhile since we're still learning the spec.

On that note, it might be worth adding a check for the spec version, since that's a thing now, and failing fast if it's unexpected. WDYT?

blowmage: Yeah, that makes sense. Are you expecting that different generator versions will support different samplegen versions? Like, will it support 1.2.0 _and_ 1.3.0? Or _only_ 1.2.0?

blowmage: Updated so it only uses 1.2.0.

jbolinger: In theory, we'll have to support multiple versions at the same time and switch based on the version as needed. In practice, I suspect that this will be years in the future, if there are ever any major revisions, and we should just stick to checking the version for now.

jbolinger:
@blowmage what else did you want to do before removing the do not merge tag? I'm good with splitting up the work to firm up the presenter abstraction and adding our own test configs as separate PRs if you are.

blowmage:
Oops, I forgot that this PR still had that label. I've added another tweak and removed the label.

gapic-generator/lib/gapic/schema/api.rb
```diff
@@ -109,17 +110,11 @@ def protoc_options
...
-            files.each do |file|
-              yaml = YAML.load_file file
-              sam += yaml["samples"] if yaml["samples"]
-            end
-          end
-          sam
+          protoc_options[:samples].to_s.split(";").flat_map do |sample_path|
+            yaml = YAML.load_file sample_path
+            # TODO: Remove use of "1.2.0" magic variable
+            yaml["samples"] if yaml["schema_version"] == "1.2.0"
```
jbolinger: Issue a warning to the console here as well?

blowmage: What should it say?

jbolinger: It doesn't really matter. I was thinking something like: `Unexpected sample file with version X found - ignoring`. 

It probably won't matter anytime soon but just so that we're not wondering why there's no output if the version does change in the future.

blowmage: Added.

jbolinger: LGTM!

PR #230, opened 2019-09-05
Add default blacklist of proto files to document

blowmage:
This change adds a blacklist of files not to generate docs for.
This mostly matches the previous GAPIC behavior, with the following
exceptions:

* Protobuf Any and Timestamp are documented when used by the API
* All Longrunning messages are documented, not just Operation


gapic-generator/templates/default/helpers/presenters/gem_presenter.rb
```diff
@@ -118,4 +120,10 @@ def gem_config key
 
     @api.configuration[:gem][key]
   end
+
+  def files_blacklist
+    gem_config(:blacklist) ||
```
blowmage: I am not 100% happy with this configuration name, but I couldn't think of anything better.

Right now this only contains proto files that should be skipped, but we could also put other values in here. For example if we wanted to skip Longrunning messages other than Operation, we could list them in the config. I don't know if we need separate lists for files and messages, so I left the name generic.

jbolinger: Right now it's for proto files so I'd go with something more explicit like `proto_file_blacklist` or something similar.

blowmage: YAML can be nested, so I would move this to `gem/blacklist/protos`, and we can add something like `gem/blacklist/messages` if we want to later.

PR #232, opened 2019-09-05
Add operation to PagedEnumerable::Page

blowmage:
This allows for users to access trailing metadata by enumerating each page of the response at a time. Paged client methods have been updated to create a PagedEnumerable object in the method instead of using a response formatter proc.

Extracted from #228.

### DIFF
--- a/gapic-common/lib/gapic/paged_enumerable.rb
+++ b/gapic-common/lib/gapic/paged_enumerable.rb
@@ -53,11 +53,12 @@ module Gapic
     # @param method_name [Symbol] The RPC method name.
     # @param request [Object] The request object.
     # @param response [Object] The response object.
+    # @param operation [GRPC::ActiveCall::Operation] the current RPC operation.
     # @param options [Gapic::CallOptions] The options for making the RPC call.
     # @param format_resource [Proc] A Proc object to format the resource object. The Proc should accept response as an
     #   argument, and return a formatted resource object. Optional.
     #
-    def initialize grpc_stub, method_name, request, response, options, format_resource: nil
+    def initialize grpc_stub, method_name, request, response, operation, options, format_resource: nil
       @grpc_stub = grpc_stub
       @method_name = method_name
       @request = request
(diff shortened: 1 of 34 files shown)


gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -192,20 +196,24 @@ def verify_response!
     #
     # @attribute [r] response
     #   @return [Object] the actual response object.
+    # @attribute [r] operation
+    #   @return [GRPC::ActiveCall::Operation] the current RPC operation.
```
jbolinger: "current" is a bit odd here. Maybe `the RPC operation used to fetch the page`?

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -192,20 +196,24 @@ def verify_response!
...
     class Page
       include Enumerable
-      attr_reader :response
+      attr_reader :response, :operation
 
       ##
       # @private
       # @param response [Object] The response object for the page.
       # @param resource_field [String] The name of the field in response which holds the resources.
+      # @param operation [GRPC::ActiveCall::Operation] the current RPC operation.
```
jbolinger: same

shared/output/cloud/showcase/lib/google/showcase/v1beta1/echo/client.rb
```diff
@@ -416,7 +420,6 @@ def wait request, options = nil, &block
                                    retry_policy: @config.retry_policy
 
             wrap_gax_operation = ->(response) { Gapic::Operation.new response, @operations_client }
-
```
jbolinger: FWIW I preferred the line break between the lamba(s) and the rpc call, but as long as it's consistent you can make the call.

blowmage: This may change soon.

gapic-common/lib/gapic/paged_enumerable.rb
blowmage: Updated.

PR #236, opened 2019-09-06
Allow call options timeout to be nil or negative

blowmage:
When timeout is nil or negative the RPC call will not specify a call deadline.
This is important for streaming calls that do not want a deadline.

### DIFF
--- a/gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
+++ b/gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
@@ -136,10 +136,14 @@ module Gapic
       private
 
       def calculate_deadline options
+        return if options.timeout.nil?
+
         Time.now + options.timeout
       end
 
       def check_retry? deadline
+        return true if deadline.nil?
+
         deadline > Time.now
       end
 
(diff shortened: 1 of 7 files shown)


gapic-common/lib/gapic/grpc/service_stub/rpc_call.rb
```diff
@@ -160,10 +160,14 @@ def compose_stream_proc stream_callback: nil, format_response: nil
       end
 
       def calculate_deadline options
+        return if options.timeout.nil?
+
         Time.now + options.timeout
       end
 
       def check_retry? deadline
+        return true if deadline.nil?
```
jbolinger: will this allow infinite retry?

blowmage: We already have infinite retry without this change. This just allows RPCs without a timeout/deadline to be made.

jbolinger: Hmm, I think we can take this offline then but I'm not sure we should allow that without a warning or opt-in.

blowmage: Do you mean you want to follow up offline after merging this? You approved so I'm not sure if I should merge this or not...

jbolinger: Yeah, sorry. This is fine to merge now.

Let's open an issue (low priority) to figure out if we want to do anything about infinite retry cases if they are a possibility. The whole retry situation is messy so, more than anything else, I just want to make sure that we document it well as I wouldn't expect most APIs to retry indefinitely by default.

PR #237, opened 2019-09-06
Add header to generated files

### LINKED ISSUE #214: Consider adding header to all generated files
Should we add a header or warning to all auto-generated files (as protoc often does)?

Context: https://github.com/googleapis/gapic-generator/issues/1986

blowmage:
The header includes the license and a generated file warning.

[fixes #214]


gapic-generator/templates/default/service/test/smoke.erb
```diff
@@ -1,10 +1,5 @@
 <%- assert_locals gem -%>
-# frozen_string_literal: true
-
-<%= render partial: "shared/license" %>
-
-# Auto-generated by gapic-generator-ruby v<%= Gapic::Generator::VERSION %> on <%= Time.now.strftime "%Y-%m-%d" %>.
```
jbolinger: Any opinions on if we should keep the timestamp in the `_warning.erb`. My personal preference is no, as you have it now, but I assume this was put here for a reason at some point?

blowmage: I removed it because it breaks the expected output comparison. If we put it back then we will need to reconsider how we test the generator's output.

jbolinger: Ok, I'm good without it (code will be tracked in vcs anyway) so let's stick with that.

PR #239, opened 2019-09-09
Add showcase tests for trailing_metadata

jbolinger:
These pass, but see the `TODO`s in streaming method tests. This might be the correct behavior, in which case we'll probably need to clearly document the flow, but I'm not sure quite yet.


shared/test/showcase/echo/chat_test.rb
```diff
@@ -66,4 +66,33 @@ def test_chat
...
+  def test_chat_with_metadata
+    options = Gapic::CallOptions.new metadata: {
+      'showcase-trailer': ["so", "much", "chat"],
+      quiet:              ["please"]
+    }
+    stream_input = Gapic::StreamInput.new
+
+    @client.chat stream_input, options do |response_enum, operation|
+      # TODO: confusing? desirable?
+      assert_nil operation.trailing_metadata
```
jbolinger: Ok to be nil here?

shared/test/showcase/echo/expand_test.rb
```diff
@@ -30,4 +30,26 @@ def test_expand
...
+    request_content = "The quick brown fox jumps over the lazy dog"
+
+    options = Gapic::CallOptions.new metadata: {
+      'showcase-trailer': ["a", "b"],
+      garbage:            ["xxx"]
+    }
+
+    @client.expand({ content: request_content }, options) do |response_enum, operation|
+      # TODO: confusing? desirable?
+      assert_nil operation.trailing_metadata
```
jbolinger: Ok to be nil here?

shared/test/showcase/echo/chat_test.rb
blowmage: I don't know how we would avoid it. The call doesn't have trailing metadata until after the call is closed.

I think this is good enough for now. If we come up with a better API then we will change this test to use it.

jbolinger: Agreed. I've opened #241 to track it and we'll defer the decision until after we've taken care of higher priority issues.

PR #240, opened 2019-09-09
Update Submodules to latest

blowmage:
Also update Speech to use the samplegen files from the googleapis repo.


blowmage:
I can see an argument for moving the commit switching to the googleapis speech samplegen files to a different PR since it isn't really part of upgrading the submodules.

blowmage:
> On a somewhat related note, was a flag ever added to the spec for filtering the samples that should be inlined in the code?

No, it hasn't been added yet, and we are not yet checking for it. We create example code for every sample we find, without checking `sample_type`.

PR #242, opened 2019-09-13
Add several Samplegen features

blowmage:
* Request options:
  * Support fields on request resource
  * Support deeply nested fields
  * Support value_is_file on fields
* Response options:
  * Support server streaming responses
  * Support pagination responses
  * Support LRO responses
  * Support loop with key/value variables (Hash)
  * Support write_file in response body


blowmage:
This does not cover client streaming methods. There doesn't seem to be any guidance on how to deal with streaming requests to the RPC in the samplegen documentation. I assume that is something being discussed internally?

gapic-generator/templates/default/service/client/method/docs/_sample.erb
```diff
@@ -2,23 +2,19 @@
 require "<%= method.service.service_require %>"
 
 client = <%= method.service.client_name_full %>.new
+<%- if sample.input_parameters.any? -%>
```
jbolinger: is this empty to force a line break?

jbolinger:
@blowmage I've asked for some more examples for streaming but don't have anything yet. I think it's a fairly low priority simply because of the number of normal methods vs. streaming. We can defer it until there's more to work off of.

gapic-generator/templates/default/service/client/method/docs/_sample.erb
blowmage: Yes

PR #243, opened 2019-09-14
Rework unit tests with stub instead of mock

blowmage:
This PR is an alternative to #226, but it uses a client stub instead of a mock to perform the yields used by the latest gapic-common gem.


gapic-generator/templates/default/service/test/method/_setup.erb
```diff
@@ -0,0 +1,39 @@
+def setup
+  @test_channel = GRPC::Core::Channel.new "localhost:8888", nil, :this_channel_is_insecure
+  @mock_stub = MiniTest::Mock.new
+  @mock_page_enum = :mock_page_enum
+  @response = :mock_response
+  @grpc_operation = :mock_grpc_operation
+  @options = {}
+end
+
+def with_stubs client: nil, paged: nil
```
blowmage: I find this method name to be a bit obtuse and not descriptive enough. I would prefer to see the stubs in the method instead.

gapic-generator/templates/default/service/test/method/_setup.erb
```diff
@@ -0,0 +1,39 @@
...
+  client ||= @mock_stub
+  paged ||= @mock_page_enum
+  Gapic::ServiceStub.stub :new, client do
+    Gapic::PagedEnumerable.stub :new, paged do
```
blowmage: I think it would be good to list the arguments, as an added check that the method is being called correctly.

gapic-generator/templates/default/service/test/method/_setup.erb
jbolinger: Why? I don't see how adding the parameters makes it any clearer. It's actually a little confusing to me (i.e. why would you always set `@mock_stub` if it's created during `setup`)? Are there test cases that wouldn't do this?

gapic-generator/templates/default/service/test/method/_setup.erb
jbolinger: For sure, this was something I hadn't gotten to in the first pass. We should add it.

gapic-generator/templates/default/service/test/method/_setup.erb
blowmage: The name and the parameters are two separate issues.

I added the parameters because I was only changing one test at a time, to get it working. Once it was working I converted the other tests. So the arguments was part of the refactoring used to get it to work. But one of the other changes is that the client stub object is created in each method, so it needs to be passed in.

The name is something else. Usually, when I see "with" I assume values are passed into the method. It seem like a method that is intended to be chained. `set_value(value).with_option(option)`. it seems to be more of a "setup" than a "with", to me at least.

>  why would you always set `@mock_stub` if it's created during `setup`

I don't think `@mock_stub` is used anymore. I think the client is always passed in.

gapic-generator/templates/default/service/test/method/_setup.erb
blowmage: This would be another reason to unfactor `with_stubs` back to the test methods.

jbolinger: Maybe. Let's get it to work for all the various tests that we'd like and then we can shuffle things around as needed.

gapic-generator/templates/default/service/test/method/_setup.erb
jbolinger: I don't really care much about the name. Feel free to use whatever name and signature you think fits here. If possible, I'd like to get this out of the generated code and into a helper but let's get it all to work first and then we can optimize, if needed, and reorganize stuff like this.

PR #245, opened 2019-09-18
Updates for grpc 1.24.0

blowmage:
This PR includes changes that are only available when using the upcoming grpc 1.24.0 release.


gapic-common/lib/gapic/grpc/status_details.rb
```diff
@@ -13,23 +13,16 @@
...
 require "grpc/google_rpc_status_utils"
-require "gapic/errors"
 require "google/protobuf/well_known_types"
 # Required in order to deserialize common error detail proto types
 require "google/rpc/error_details_pb"
 
-module Gapic
-  class GapicError < StandardError
+module GRPC
+  class BadStatus < StandardError
```
jbolinger: I suppose this could be considered out of scope for this PR, but we should add docs to this method. Can we at least `TODO` for a doc pass later?

gapic-common/gapic-common.gemspec
```diff
@@ -39,7 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "googleapis-common-protos", ">= 1.3.9", "< 2.0"
   spec.add_dependency "googleapis-common-protos-types", ">= 1.0.4", "< 2.0"
   spec.add_dependency "googleauth", ">= 0.6.2", "< 0.10.0"
-  spec.add_dependency "grpc", ">= 1.7.2", "< 2.0"
+  spec.add_dependency "grpc", "1.24.0.pre1"
```
jbolinger: I assume we'll go back to the range when it's no longer a pre-release? Do you have an ETA?

gapic-common/lib/gapic/paged_enumerable.rb
```diff
@@ -95,7 +95,7 @@ def each
     #
     # @yield [Page] Gives the pages in the stream.
     #
-    # @raise [GapicError] if it's not started yet.
+    # @raise if it's not started yet.
```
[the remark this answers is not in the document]
blowmage: I don't know where this error is coming from. I think it might have been copied over from Gax, but I dunno. I'm not sure what type of error this is, if it even exists.

gapic-common/gapic-common.gemspec
blowmage: It usually takes a couple weeks for grpc to go from pre to actual release. By the end of the month?

gapic-common/lib/gapic/paged_enumerable.rb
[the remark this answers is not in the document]
jbolinger: Ah, if it was already wrong and we don't know maybe put another `TODO` so we catch it during the doc pass? It would be good to document the error types since this is used for many things.

blowmage:
Closing in favor of #246

PR #285, opened 2019-12-23
Full Gem Output

blowmage:
Add full gem to expected output. This will allow us to run `bundle install && bundle exec rake ci` on the output.


shared/Gemfile
```diff
@@ -4,7 +4,9 @@ source "https://rubygems.org"
 
 gem "gapic-common", path: "../gapic-common"
 gem "gapic-generator", path: "../gapic-generator"
-gem "gapic-generator-cloud", path: "../gapic-generator-cloud"
+# gem "gapic-generator-ads", path: "../gapic-generator-ads"
+# gem "gapic-generator-cloud", path: "../gapic-generator-cloud"
```
quartzmo: Why are these lines commented out?

quartzmo:
After running `bundle exec rake gen` (and `ci`):

```sh
$ cd shared/output/gems/cloud/speech/
Chriss-iMac:speech quartzmo$ bundle update
Fetching https://github.com/googleapis/gapic-generator-ruby.git
error: could not lock config file .git/modules/shared/api-common-protos/config: No such file or directory
warning: Could not unset core.worktree setting in submodule 'shared/api-common-protos'
error: could not lock config file .git/modules/shared/googleapis/config: No such file or directory
warning: Could not unset core.worktree setting in submodule 'shared/googleapis'
error: could not lock config file .git/modules/shared/protobuf/config: No such file or directory
warning: Could not unset core.worktree setting in submodule 'shared/protobuf'
The path `/Users/quartzmo/code/google/codez/gapic-generator-ruby/shared/output/gems/cloud/google-cloud-core` does not exist.
```

quartzmo:
The following lines in the generated gems' `Gemfile`s prevent `bundle` from succeeding, because the gems are not located in the `google-cloud-ruby` repo:

```sh
gem "google-cloud-core", path: "../google-cloud-core"
gem "google-cloud-env", path: "../google-cloud-env"
```

shared/Gemfile
blowmage: meant to be removed.

quartzmo:
Despite the error, I think we should leave those relative requires (We should continue to presume that cloud gems will live in the google-cloud-ruby monorepo.)

PR #461, opened 2020-06-04
[gg-ruby] docs: add Ruby and Bundler set-up steps to README

miraleung:


viacheslav-rostovtsev:
The resulting file has some issues when I look at it in the preview window -- e.g. dangling ```sh instructions, some lines disappear, etc. Could be a difference in your editor's markdown interpretation if you are using any. 
I think that having a simple "getting started" instruction *for non-Ruby developers* is a thing worth adding to Readme, requesting @dazuma 's opinion on whether `apt-get install gem`/`brew install gem` is a ruby-community approved way for that.

miraleung:
Fixed, reverted the auto-format changes.

PR #472, opened 2020-06-23
laundry list of Penelope's changes for review

viacheslav-rostovtsev:

PR #473, opened 2020-06-23
fix: removes bundler references from Gemfiles and gemspecs

viacheslav-rostovtsev:


viacheslav-rostovtsev:
includes https://github.com/googleapis/gapic-generator-ruby/pull/472#issuecomment-648435314

PR #494, opened 2020-07-23
Make gapic-generator-ads a package and export its config.yml file.

aohren:
This exports the ads `config.yml` file so that it can be referenced by a `ruby_gapic_ads_library` rules in another project (specifically, the googleads project's build files).


viacheslav-rostovtsev:
we're going another route with this as discussed offline; the yml config should be replaced by the command line params

PR #510, opened 2020-09-10
chore: add CODEOWNERS and LICENSE

JustinBeckwith:


dazuma (approved):
Each of the package directories has a LICENSE file already, but I guess it doesn't hurt to have one at the top level as well.

PR #547, opened 2021-02-16
chore: remove nonbundler rules

viacheslav-rostovtsev:
all prod toolchains are on bundler, nonbundler rules are legacy and due for removal


vam-google (approved):
LGTM, but please confirm that with these removed, all googleapis repository still builds fine.
Also, please specify in the PR description why it is needed before and why it is safe to remove now.

dazuma (approved):
stamp

PR #548, opened 2021-02-16
feat: add showcase library to the noncloud examples

viacheslav-rostovtsev:
noncloud example can benefit from having a real library in it, now that showcase Bazel bindings are there


vam-google (approved):
LGTM

dazuma (approved):

Issue #539, opened 2021-01-29
bazel: move description & title protoc options to attributes


noahdietz:
If the values of `ruby-cloud-description` or `ruby-cloud-title` have spaces in them (which is highly likely) and these options are supplied to `extra_protoc_parameters` in a `BUILD.bazel`, re-generation (via `build_gen`) of an existing `BUILD.bazel` file results in the values of `ruby-cloud-description` and `ruby-cloud-title` being expanded and supplied as `extra_protoc_parameters` which breaks the build. 

[Example BUILD.bazel](https://github.com/googleapis/googleapis/blob/c71a9c4ff7bc5179f2e94af31bc95ce4922575ff/google/cloud/channel/v1/BUILD.bazel#L289-L290). From googleapis, reproduce the issue by running  `bazel run //:build_gen -- --src=google/cloud/channel/v1` and attempt to build the `ruby_gapic_library` target: `bazel build //google/cloud/channel/v1:channel_ruby_gapic`.

While this doesn't sound like an issue gapic-generator-ruby should deal with, it is infeasible to fix in `build_gen` at the moment, because the `buildozer` API for merging `BUILD.bazel` files does not support `List` attributes (like `extra_protoc_parameters`) very well (it just puts all the items into a single, square brace-wrapped, space-delimited string).

So, this is a request to add a `description` and `title` attribute to `ruby_gapic_library` that would subsequently be added to generator invocation as `ruby-cloud-description={value}` and `ruby-cloud-title={value}"`. We could then have `build_gen` preserve these individual attributes on `ruby_gapic_library`, which for singular `String` attributes works much better.

WDYT?

cc @alexander-fenster

viacheslav-rostovtsev:
This looks like a reasonable straightforward thing to do.

PR #549, opened 2021-02-17
feat: add additional params for Ruby-Bazel cloud rules

viacheslav-rostovtsev:
Ruby-Bazel cloud rule needs to move gem-title and gem-description into additional parameters because build_gen cannot handle arrays with strings with spaces. This provides
see #539 for details


rules_ruby_gapic/ruby_gapic.bzl
```diff
@@ -56,17 +56,40 @@ def ruby_gapic_library(
...
   srcs,
+  ruby_cloud_title = "",
+  ruby_cloud_description = "",
   extra_protoc_parameters = [],
   grpc_service_config = None,
   **kwargs):
   
+  if extra_protoc_parameters:
+    for key_val_string in extra_protoc_parameters:
+        key_val_split = key_val_string.split("=", 1)
```
vam-google, line 78: do you want to trim the key/value pairs as well (to accommodate for leading and trailing whitespaces)?

vam-google (approved):
LGTM, thought it still feels like there must be a way to have spaces in `extra_protoc_parameters` (spaces seems like a fundamental thing and buildozer not supporting them seems too poor job for buildozer to be true).

dazuma (approved):

rules_ruby_gapic/ruby_gapic.bzl
viacheslav-rostovtsev: that's a good idea

PR #559, opened 2021-02-24
chore: add yardopts-cloudrad to CloudGenerator#generate

TheRoyalTnetennba:


dazuma (approved):

dazuma (commented):
You probably need to regenerate and check in all the generated files, in order to get tests to pass.

PR #568, opened 2021-03-13
feat: add parameter to allow overriding of wrapper gems in cloud

viacheslav-rostovtsev:


dazuma (approved):
LGTM. And thanks for setting up the separate gen tasks. I've wanted something like that for a long time.

Issue #570, opened 2021-03-22
Gapic metadata (drift manifest) generation is incorrect


viacheslav-rostovtsev:
Spec at https://github.com/googleapis/googleapis/blob/d6b4fb337caf6eccb606728ef727ac76364a4f14/gapic/metadata/gapic_metadata.proto#L79 shows that the key for service should be `rpcs` and not `methods`
Spec at https://github.com/googleapis/googleapis/blob/d6b4fb337caf6eccb606728ef727ac76364a4f14/gapic/metadata/gapic_metadata.proto#L89 shows that the list of generated method names should be in a hash with key `methods` instead of directly in the `name-of-rpc` value.

PR #571, opened 2021-03-22
feat: add CLI guard for drift manifest generation, fix: incorrect structure

viacheslav-rostovtsev:
fix: incorrect structure for drift manifest 

fixes https://github.com/googleapis/gapic-generator-ruby/issues/569
fixes https://github.com/googleapis/gapic-generator-ruby/issues/570


gapic-generator/test/gapic/presenters/package/showcase_test.rb
```diff
@@ -54,9 +54,9 @@ def test_drift_showcase_v1beta1
     refute_nil echo_gprc
 
     assert_equal "::Google::Showcase::V1beta1::Echo::Client", echo_gprc[:libraryClient]
-    assert_equal 7, echo_gprc[:methods].length
+    assert_equal 7, echo_gprc[:rpcs].length
```
vchudnov-g, line 57: Did you mean `echo_gprc` to be `echo grpc`?

gapic-generator/lib/gapic/presenters/method_presenter.rb
```diff
@@ -232,6 +232,19 @@ def grpc_method_name
...
+      ##
+      # Returns a hash with a drift_manifest of this rpc method
+      # describing correspondence between the proto description
+      # of the rpc with the generated code for the method.
+      # For ruby currently [03/2021] only one method is generated per RPC,
+      # so the correspondence is very basic.
+      # See https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto
+      #
+      # @return [Hash]
+      def drift_manifest
```
vchudnov-g, line 244: Is it possible for you to populate the `gapic_metadata.proto` and then call a standard proto-to-JSON serializer on the proto? This would avoid discrepancies like the one you're fixing, and would be useful should we add new features in the future.

viacheslav-rostovtsev: it's not possible to use the ruby classes generated by the standard proto-to-JSON serializer inside the ruby microgenerator at the moment

gapic-generator/test/gapic/presenters/package/showcase_test.rb
viacheslav-rostovtsev: `echo_grpc` is a variable declared on the line 52 so unless I misunderstand the question, no, I meant `echo_grpc`.

vchudnov-g: Sorry, my typo obscure the fact I was referring to your apparent typo: you say `gprc` instead of `grpc` (you have two letters transposed).

gapic-generator/lib/gapic/presenters/method_presenter.rb
vchudnov-g: So there is no standard code to take a proto populated by Ruby and emit JSON? Bummer.

dazuma (approved):

viacheslav-rostovtsev:
I've made a change to enable it by default in gg-cloud, since otherwise we'll have to enable it in every build.bazel and/or synth.py

PR #584, opened 2021-05-14
fix: Upload prebuilt ruby binary which works on workstations and CI image

vam-google:


vam-google:
@viacheslav-rostovtsev PTAL

alexander-fenster (approved):

viacheslav-rostovtsev (approved):

PR #598, opened 2021-06-08
fix(bazel): Replace monolith deps with rules_gapic [ruby]

miraleung:
This takes us closer to removing the monolith from googleapis entirely - it is needed for the preliminary removal of the monolith rules from googleapis' `switched_rules_by_language`.
Confirming that `bazel_example` builds with this PR.


vam-google (approved):
LGTM

viacheslav-rostovtsev (approved):
LGTM

PR #634, opened 2021-07-26
docs: Add test and linting steps to README [ggr]

miraleung:


README.md
```diff
@@ -213,6 +213,26 @@ Contributions to this library are always welcome and highly encouraged.
...
+To run all tests:
+
+```sh
+$ bundle exec rake test
+```
+
+To test a single file:
+
+```sh
+$ bundle exec rake test TEST=path/to/test/file.rb
```
viacheslav-rostovtsev, line 228: This does not work from the root

viacheslav-rostovtsev (changes requested):
If we are to add a developer guide, it needs to cover more ground, mentioning at least the generator structure and describing the development workflow, e.g. protos vs bin, unit tests vs gold tests, vs unit-tests-on-gold vs showcase. 

As it is, it's only useful to a very narrow scope of potential developers, not to mention being wrong, e.g. 
`bundle exec rake test TEST=path/to/test/file.rb` does not do what this says when run from the root of the project, and the distinction of root vs subgenerator is not described in the guide.

README.md
miraleung: Fixed.

miraleung:
This PR is meant to incrementally add developer steps that would aid in setting up and working with this codebase. A full developer guide could come as a result of similar accumulated incremental changes, or this PR could make such an effort easier.

viacheslav-rostovtsev (approved):

PR #828, opened 2022-09-22
feat: add the error behavior

viacheslav-rostovtsev:


gapic-common/lib/gapic/rest/server_stream.rb
```diff
@@ -50,9 +51,16 @@ def each
           while @ready_objs.length.zero?
             begin
               chunk = @json_enumerator.next
-              return unless chunk
+              next unless chunk
```
atulep, line 54: Why next instead of return ?

gapic-common/lib/gapic/rest/threaded_enumerator.rb
```diff
@@ -38,16 +37,26 @@ def initialize &block
...
       end
 
       def next
         @in_q.enq :next
         chunk = @out_q.deq
-        raise StopIteration if chunk.nil?
+
+        if chunk.is_a? StandardError
+          @out_q.close
+          @in_q.close
```
atulep, line 51: How about also clearing the out_q ? It can be misleading receiving an error about dangling content and having old stuff in the queue.

atulep (approved):

PR #845, opened 2022-11-04
fix: deadlock fix

atulep:
Fixes a bug in the threaded enumerator where out_q indefinitely waits for nil (terminating) value.


viacheslav-rostovtsev (approved):
LGTM

atulep:
@vam-google fyi

dazuma (approved):

PR #888, opened 2022-12-16
fix: fail when rest transport specified but no HTTP bindings

viacheslav-rostovtsev:
fixes [887](https://github.com/googleapis/gapic-generator-ruby/issues/887)


dazuma (approved):
So we want a hard failure here (e.g. a TAP presubmit failure) rather than raise a warning and disable Rest transport? If so, LGTM.

PR #897, opened 2023-01-09
fix: remove a deprecated method usage that new Ruby does not like

viacheslav-rostovtsev:
NB: will work after rebasing on https://github.com/googleapis/gapic-generator-ruby/pull/896


dazuma (approved):
Use `fix:` for the conventional commit message.

PR #1037, opened 2024-01-26
chore: update showcase dir + submodules

aandreassa:
Ran `git submodule update --remote`, `toys bin` and `toys gen`.

Updated googleads to v15 due to error: `Could not make proto path relative: google/ads/googleads/v12/services/campaign_service.proto: No such file or directory`


shared/protos/google/showcase/v1beta1/echo.proto
aandreassa, line 1: This directory was copied from [gapic-showcase](https://github.com/googleapis/gapic-showcase/tree/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/schema/google/showcase/v1beta1)

PR #1071, opened 2024-04-25
feat: Remove api-common-protos dependency

aandreassa:


dazuma (approved):
Looking great! Thanks!

Issue #1090, opened 2024-07-09
Resolve reserved keyword conflict with variable names in path patterns


aandreassa:
https://github.com/googleapis/google-cloud-ruby/pull/26358 introduced a path variable `module`, which conflicts with reserved keywords, failing generation.

Current plan is:
1. Introduce an owlbot rule to affected library and unblock generation (adding sufix `_param`).
2. Add a fix to the generator itself, where keyword names will be properly escaped.

Note: `google-cloud-support-v2/.owlbot.rb` already contains a similar patch and should be tested on upcoming change. In that example, the word `case` gets converted to `ccase`.

For reference, PiperOrigin-RevId: 650374217

Pointers: https://github.com/googleapis/gapic-generator-ruby/pull/1086

aandreassa:
#1101 should address the last issue we had on naming the local variables. Once the pending PR's are in, I can get started on the generator release to close this out.

PR #1096, opened 2024-08-02
feat: Resolve keyword conflicts with path pattern variable names

aandreassa:
More info: https://github.com/googleapis/gapic-generator-ruby/issues/1090


dazuma (approved):
Looks great! Thanks!

PR #1101, opened 2024-08-06
feat: Prevent keyword collision with field names

aandreassa:
Addresses the last part of #1090, preventing local variables from having conflicts with Ruby reserved keywords.


dazuma (approved):
Great, thanks!

PR #1110, opened 2024-08-22
feat: Support local binding for variables with keyword collision

aandreassa:
Reverting changes from https://github.com/googleapis/gapic-generator-ruby/pull/1096 and adding `binding.local_variable_get` for collisions from the generator.


dazuma (approved):
LGTM

PR #1169, opened 2025-04-24
feat: bless "::Google::Auth::BaseClient" credentials type

viacheslav-rostovtsev:
Set "::Google::Auth::BaseClient" credentials type as an allowed by default credentials type in Config


dazuma (approved):
LGTM

PR #1207, opened 2025-06-23
fix: docs for oneof parameters

viacheslav-rostovtsev:
fixes: 1149


dazuma (changes requested):
The challenge here is that we want the updated language, describing these as "parameters", to show up only where the fields are actually being documented as method parameters (e.g. here: https://github.com/googleapis/gapic-generator-ruby/pull/1207/files#diff-9ef7164c65316d68bc5179840ab4c5da051d5e5f176206363263138eaadc1551) but not when they are documented as fields in an object (e.g. here: https://github.com/googleapis/gapic-generator-ruby/pull/1207/files#diff-84a7d1c30e99d50b785d3bdc1cd1d47e1c3eeb299afda05132d80642ca7a40b4).

dazuma (approved):
Thanks!

PR #1209, opened 2025-06-26
fix: remove require when rest operations are not generated

viacheslav-rostovtsev:
In cases when LRO methods are present but don't have `google.api.http` annotations the rest operations mixin is not generated. However the require was still emitted for it. This rectifies.


dazuma (approved):
LGTM, thanks!

PR #1214, opened 2025-08-12
fix: Remove unnecessary services import from tests

aandreassa:
This is a fix to remove manual changes from https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-pubsub-v1/.owlbot.rb


aandreassa:
@viacheslav-rostovtsev It fails bc it conflicts with the service renaming we did for PubSub modernization. https://github.com/googleapis/google-cloud-ruby/blob/eeaca40a5933d59c9449baaa75e64d5d9f57ea9b/google-cloud-pubsub-v1/.repo-metadata.json#L19

PR #1279, opened 2026-04-10
feat: Skip generation for services without RPC methods defined

aandreassa:
Example of empty service: [google-cloud-dataplex-v1](https://github.com/googleapis/googleapis/blob/db5d0d9e1a97acc63b787a67cd946e8244b77c88/google/cloud/dataplex/v1/content.proto#L30-L33)

fix: safely handle bindings_override passing in local LRO clients

When an API includes custom standalone LRO clients alongside ones using mixins like Location, the generator tries to configure those mixins inside the LRO client itself.

Since these LRO clients don't support the `bindings_override` attribute, this causes crashes with`NoMethodError`. This change adds a simple check to ensure we only try to pass that setting when the parent client actually defines it.


aandreassa:
I tested this locally in `google-cloud-dataform-v1beta1` using a local build of `gapic-generator-ruby` in `googleapis` . Here is the diff (expected line was deleted):

```
$ git diff
diff --git a/google-cloud-dataform-v1beta1/lib/google/longrunning/operations/rest/client.rb b/google-cloud-dataform-v1beta1/lib/google/longrunning/operations/rest/client.rb
index f0ddf1e72e..d150c41ef1 100644
--- a/google-cloud-dataform-v1beta1/lib/google/longrunning/operations/rest/client.rb
+++ b/google-cloud-dataform-v1beta1/lib/google/longrunning/operations/rest/client.rb
@@ -182,7 +182,6 @@ module Google
               config.quota_project = @quota_project_id
               config.endpoint = @operations_stub.endpoint
               config.universe_domain = @operations_stub.universe_domain
-              config.bindings_override = @config.bindings_override
               config.logger = @operations_stub.logger if config.respond_to? :logger=
             end

```

viacheslav-rostovtsev (approved):
LGTM

quartzmo (approved):
LGTM

PR #1285, opened 2026-05-07
chore: Add Ruby 4.0 to CI

aandreassa:
chore: Update README.md comments

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,11 +16,16 @@ jobs:
   tests:
     if: ${{ github.repository == 'googleapis/gapic-generator-ruby' }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: ["3.2", "4.0"]
+    # Allow failures on 4.0; generator primarily runs on 3.2.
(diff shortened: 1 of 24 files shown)


.github/workflows/ci.yml
```diff
@@ -16,11 +16,16 @@ jobs:
   tests:
     if: ${{ github.repository == 'googleapis/gapic-generator-ruby' }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: ["3.2", "4.0"]
+    # Allow failures on 4.0; generator primarily runs on 3.2.
```
quartzmo, line 22: This is interesting, can you explain a bit more? (Update the comment if you think there's more context we can provide for maintainers, or if not, maybe just add some detail here in the PR comments for me.)

aandreassa: IMO we shouldn't go as far as blocking CI on 4.0 right now, but we should be aware if there are errors 
(locally we are usually using the latest Ruby version).

Historically, this has just been running on the minimum version (3.2), so it's a good first step to get signals.

quartzmo (approved):

PR #1286, opened 2026-05-13
fix: rename variable in generated test methods to avoid field name collision

aandreassa:
Renames the local `client` variable (used when instantiating the service client under test) to `client_stub` across all gRPC and REST test templates.

This prevents variable shadowing and collisions with existing names.

This fixes a generation error in `google-cloud-run-v2` where the gem introduces a field named `client` in [`SubmitBuildRequest`](https://github.com/googleapis/googleapis/blob/a0cedfb6af9305b49db07dccaa7d0c8985e07e70/google/cloud/run/v2/build.proto#L143).

Failed generation PR: https://github.com/googleapis/google-cloud-ruby/pull/32252


quartzmo (commented):
Do you think it's too risky to name this variable `c` instead of `client_stub`?

It would save 10 characters!

quartzmo:
The Go clients use `c`. For example: https://github.com/googleapis/google-cloud-go/blob/main/accessapproval/apiv1/access_approval_client.go#L272

quartzmo (approved):

PR #1309, opened 2026-07-08
feat: Add minitest-reporters dependency and setup to generator templates

quartzmo:
Write minitest execution output to JUnit XML reports. This enables centralized test result indexing, flake tracing, and build health metrics reporting inside TestGrid dashboards.


gapic-generator/templates/default/gem/test_helper.text.erb
```diff
@@ -1,6 +1,31 @@
...
 <%= render partial: "shared/header" -%>
 
+# Configure JUnit XML test formatting for TestGrid build health and test failure tracking.
+if ENV["CI"] || ENV["KOKORO_JOB_NAME"]
+  begin
+    require "fileutils"
+    FileUtils.mkdir_p "tmp/reports"
+    require "minitest/reporters"
+    unless defined? SpongeReporter
+      class SpongeReporter < Minitest::Reporters::JUnitReporter
```
torreypayne: We define `class SpongeReporter` roughly 18 times in this PR, and the implementation is nearly identical across this PR and the other 2 for the other repos; is there any way we could DRY this out?

aandreassa: This should be fine, @torreypayne! All other `SpongeReporter` definitions are under `shared/output`, which is an auto-generated directory we use for testing gems!

aandreassa (approved):

aandreassa:
🎉

gapic-generator/templates/default/gem/test_helper.text.erb
quartzmo: Working on it!

quartzmo: Replaced the boilerplate with:

```
# Load JUnit XML formatter from googleapis/ruby-common-tools to write
# tmp/reports/sponge_log.xml for Kokoro/TestGrid build health tracking.
if ENV["CI"] || ENV["KOKORO_JOB_NAME"]
  begin
    require "gapic/minitest_junit_preloader"
  rescue LoadError
  end
end
```

which loads the monkeypatch from https://github.com/googleapis/ruby-common-tools/pull/429

aandreassa (approved):

PR #1319, opened 2026-07-21
fix(gapic-generator): Indent note in the Config#credentials attribute properly

dazuma:
Fixes the indentation on a multi-line `@note` yard tag. Without proper indentation, only the first line gets included in the note, while the remaining lines are orphaned into the general description of the method. (For an example of what this looks like, see https://rubydoc.info/gems/google-cloud-secret_manager-v1/Google/Cloud/SecretManager/V1/SecretManagerService/Client/Configuration#credentials-instance_method — the second warning note on the credentials method is cut off mid-sentence, and the remainder of the note appears below just above the examples.)

I found this as I was using secret_manager-v1 as a test case in a YARD plugin I'm working on.

Hope y'all are doing well!

### DIFF
--- a/gapic-generator/templates/default/service/client/_config.text.erb
+++ b/gapic-generator/templates/default/service/client/_config.text.erb
@@ -65,11 +65,11 @@
 #     end
 #
 #   @note Warning: If you accept a credential configuration (JSON file or Hash) from an
-#   external source for authentication to Google Cloud, you must validate it before
-#   providing it to a Google API client library. Providing an unvalidated credential
-#   configuration to Google APIs can compromise the security of your systems and data.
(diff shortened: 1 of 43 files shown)


aandreassa:
/gcbrun

Thanks @dazuma, we appreciate the contribution! We will likely hold-off until next week at least, since we have a couple changes in-flight. Adding do not merge for now!

dazuma:
> /gcbrun
> 
> Thanks @dazuma, we appreciate the contribution! We will likely hold-off until next week at least, since we have a couple changes in-flight. Adding do not merge for now!

@aandreassa No worries. This would trigger a regen of every gapic, so y'all would want to manage it carefully.

PR #1322, opened 2026-07-24
fix: generate double-wildcard path patterns with capture groups

quartzmo:


aandreassa (changes requested):
Need to update goldens?

aandreassa (approved):
LGTM!

PR #1327, opened 2026-07-31
fix: upgrade minitest-reporters dependency to ~> 1.8.0 in wrapper gem template

quartzmo:
refs: #1316


aandreassa:
Ah, we totally missed this. I can integrate this with my other generator changes. TY!

cc @torreypayne  https://github.com/googleapis/google-cloud-ruby/blob/a17b0e13ba374c20811dd278dcdf7784c97709a4/google-cloud-shell/Gemfile.lock#L141

aandreassa (approved):

PR #1331, opened 2026-08-07
test: run showcase integration tests over TLS

torreypayne:
Starts the local `gapic-showcase` server with `--tls` and routes every gRPC and REST test client through the encrypted endpoint. The suite previously connected over plaintext, so no test here ever exercised a real TLS handshake. No PQC assertions yet — this only establishes the transport #1351 asserts against.

Showcase mints its own certificate under `--tls` and publishes the signing CA through `--ca-cert-output-file`, so the harness runs no certificate authority of its own.

### Reviewer notes

- **Trust is wired per server**, because each `--tls` server mints its own CA. gRPC gets explicit `ChannelCredentials`, since the C-core resolves `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` once per process and cannot express a per-server root. REST uses `SSL_CERT_FILE`, which `Net::HTTP` re-reads per connection. Both are restored on teardown.
- **The readiness wait targets the CA file, not the port.** `endpoint.go` binds before generating TLS material: over 40 boots on four cores the port accepted at ~0.01s, the CA at a median 1.4s, max 3.2s. A port probe reports ready seconds early, the race that flaked Rust ([#6542](https://github.com/googleapis/google-cloud-rust/issues/6542), fixed in [#6547](https://github.com/googleapis/google-cloud-rust/pull/6547)). A fixed sleep is no substitute: 5% of boots need over 3s, and the suite starts three servers. It parses the file, since existence is briefly true mid-write.
- **`spawn_showcase` reaps a child whose startup failed.** Otherwise it survives holding `:7469`, the next run reuses it, and every TLS test fails opaquely.

---
Sits on #1330 (fork-hosted, so GitHub excludes it from the stack) · **Companion**: [ruby-core-libraries#73](https://github.com/googleapis/ruby-core-libraries/pull/73) raises the `gapic-common` grpc floor.
Design: [go/client-libraries:ruby-pqc](http://goto.google.com/client-libraries:ruby-pqc) · Parent: [go/cloudsdk-pqc-ruby](http://goto.google.com/cloudsdk-pqc-ruby)

### DIFF
--- a/shared/Gemfile
+++ b/shared/Gemfile
@@ -17,3 +17,4 @@ gem "minitest-autotest", "~> 1.0"
 gem "minitest-focus", "~> 1.0"
 gem "ostruct", "0.6.3"
 gem "rake", ">= 12.0"
+gem 'fiddle'
--- a/shared/test/showcase/pqc_test.rb
+++ b/shared/test/showcase/pqc_test.rb
@@ -1,0 +1,44 @@
+require "test_helper"
+require "fiddle"
+require "openssl"
+
+module PQProbe
+  LIBSSL = Fiddle.dlopen("libssl.so.3")
+  GET_GROUP_NAME = Fiddle::Function.new(
+    LIBSSL["SSL_get0_group_name"],
+    [Fiddle::TYPE_VOIDP], Fiddle::TYPE_VOIDP
+  )
+  DATA_OFFSET = 32
+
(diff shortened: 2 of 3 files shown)


amanda-tarafa (approved):
LGTM as far as my understanding goes.

Still would like @quartzmo eyes on it.

PR #1351, opened 2026-09-15
test: add post-quantum cryptography conformance tests for showcase

torreypayne:
Adds `shared/test/showcase/pqc_test.rb`, proving Ruby clients negotiate `X25519MLKEM768` against Showcase over both transports.

Ruby never performs the key exchange itself — gRPC delegates to BoringSSL vendored in the `grpc` gem, REST to system OpenSSL. The suite therefore asserts on what the *server* observed, via the `x-showcase-tls-group` and `x-showcase-tls-client-supported-groups` metadata Showcase reflects onto every response. [Python](https://github.com/googleapis/google-cloud-python/blob/main/packages/gapic-generator/tests/system/test_pqc.py), [Node](https://github.com/googleapis/google-cloud-node/blob/main/core/packages/gax/test/test-application/src/pqc-test.ts), [Java](https://github.com/googleapis/google-cloud-java/blob/main/java-showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITPostQuantumCryptography.java), [C#](https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs), and C++ assert the same way.

Two scenarios per transport: default negotiation, and fallback against a classical-only server.

### Reviewer notes

- **The gRPC tests assert `grpc >= 1.83` up front.** Nothing here pins `grpc` — it resolves through `gapic-common`, so PQC holds only because bundler picks the newest match. Without the check a downgrade reads as a protocol bug. [ruby-core-libraries#73](https://github.com/googleapis/ruby-core-libraries/pull/73) raises that floor.
- **The fallback tests also assert the client still *offered* `X25519MLKEM768`.** Otherwise the test would pass against a client that had lost PQC entirely. For REST that half is gated on `SHOWCASE_REQUIRE_REST_PQC`, since below OpenSSL 3.5 there is no PQC group to withhold.
- **REST never skips.** No runner available to us has OpenSSL >= 3.5, so a version-guarded `skip` would mean REST is never exercised at all. It runs against a `{X25519MLKEM768, X25519}` allowlist; #1352 tightens that on a 3.5 image.
- `normalize_headers` fails loudly when TLS metadata is absent, so plaintext cannot silently no-op the suite.

Green, zero skips across OpenSSL 3.0.18 / 3.5.5 / 3.6.3 and Ruby 3.2 / 4.0.

---
Sits on #1330 (fork-hosted, so excluded from the stack).
Design: [go/client-libraries:ruby-pqc](http://goto.google.com/client-libraries:ruby-pqc) · Parent: [go/cloudsdk-pqc-ruby](http://goto.google.com/cloudsdk-pqc-ruby)


shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+#
+class PqcTest < ShowcaseTest
+  # Header carrying the key exchange group selected during the handshake.
+  NEGOTIATED_GROUP_HEADER = "x-showcase-tls-group"
+
+  # Header carrying every key exchange group the client offered in ClientHello.
+  CLIENT_GROUPS_HEADER = "x-showcase-tls-client-supported-groups"
+
+  # The hybrid post-quantum group both Go and BoringSSL prefer by default.
+  PQC_GROUP = "X25519MLKEM768"
```
amanda-tarafa: @robertvoinescu-work did something in .NET, which was look for the substring MLKEM to cover more bases here and not have to update as often.

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+  REST_OPENSSL_SUPPORTS_PQC =
+    Gem::Version.new(OpenSSL::OPENSSL_LIBRARY_VERSION.split[1]) >= MINIMUM_REST_OPENSSL_VERSION
+
+  def test_grpc_negotiates_post_quantum_key_exchange
+    assert_grpc_pqc_capable
+    headers = grpc_tls_headers new_echo_client
+
+    assert_equal PQC_GROUP, headers[NEGOTIATED_GROUP_HEADER],
+                 "gRPC handshake did not negotiate post-quantum key exchange"
+    assert_includes offered_groups(headers), PQC_GROUP,
```
amanda-tarafa: Wondering if these two asserts should be inverted in order. First we check what the client advertised, then we check whether PQC was selected.

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+    with_showcase_tls_groups CLASSICAL_ONLY_CODEPOINTS do |port, ca_path|
+      headers = grpc_tls_headers grpc_echo_client_for(port, ca_path)
+
+      assert_equal CLASSICAL_GROUP, headers[NEGOTIATED_GROUP_HEADER],
+                   "gRPC client failed to fall back to classical key exchange"
+      # Negotiating X25519 alone proves nothing: a client that had lost
+      # post-quantum support entirely would produce the same result. What is
+      # being tested is that a PQC-capable client still interoperates with a
+      # classical-only server.
+      assert_includes offered_groups(headers), PQC_GROUP,
```
amanda-tarafa: Same note as before, interchange the order of these asserts.

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+
+      assert_equal CLASSICAL_GROUP, headers[NEGOTIATED_GROUP_HEADER],
+                   "REST client failed to fall back to classical key exchange"
+      # Same reasoning as the gRPC case, but only checkable where the host
+      # OpenSSL implements ML-KEM at all; below 3.5 the client has no
+      # post-quantum group to withhold, so there is no fallback to observe.
+      # Gated on capability rather than on REQUIRE_REST_PQC so that a
+      # PQC-capable host runs the real assertion even when strict mode is off.
+      if REST_OPENSSL_SUPPORTS_PQC
+        assert_includes offered_groups(headers), PQC_GROUP,
```
amanda-tarafa: Invert order of asserts

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+  #
+  # The suite is not parallelized, so swapping a process-wide environment
+  # variable for the duration of the block is safe; adding parallelize_me! to
+  # this file would break that assumption.
+  #
+  # @param codepoints [String] Comma separated IANA key exchange group IDs.
+  # @yieldparam port [Integer]
+  # @yieldparam ca_path [String]
+  # @return [void]
+  def with_showcase_tls_groups codepoints
```
amanda-tarafa: Shouldn't this be on the test helper file?

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+  #
+  # Every server started with --tls mints its own certificate authority, so the
+  # auxiliary server cannot share a trust root with the main harness. Both
+  # transports have to be pointed at the CA yielded here: REST through
+  # SSL_CERT_FILE, which is scoped to the block below because Net::HTTP rebuilds
+  # its trust store per connection, and gRPC through explicit credentials built
+  # from the yielded path.
+  #
+  # The suite is not parallelized, so swapping a process-wide environment
+  # variable for the duration of the block is safe; adding parallelize_me! to
```
amanda-tarafa: Can you assert that parallelize_me has not been added?

shared/test/showcase/pqc_test.rb
```diff
@@ -0,0 +1,306 @@
...
+  # assertion, and is used by the CI job that runs on an image pinned to
+  # OpenSSL >= 3.5.
+  #
+  # @param headers [Hash{String=>String}]
+  # @return [void]
+  def assert_rest_key_exchange headers
+    negotiated = headers[NEGOTIATED_GROUP_HEADER]
+    offered = offered_groups headers
+
+    if REQUIRE_REST_PQC
```
amanda-tarafa: I'd use this variables differently and that will simplify this:

- Assert that if `REQUIRES_REST_PQC` is true then `REST_OPENSSL_SUPPORTS_PQC` is true.
- Assert PQC when `REST_OPENSSL_SUPPORTS_PQC` is true.
- Assert classical when `REST_OPENSSL_SUPPORTS_PQC` is false.

amanda-tarafa (changes requested):
Mostly good, just some improvements I think.

shared/test/showcase/pqc_test.rb
torreypayne: Addressed your feedback and updated the diff accordingly; exactly one outcome passes now, either PQC is required and supported, or it's not.

shared/test/showcase/test_helper.rb
```diff
@@ -173,6 +173,79 @@ def new_compliance_rest_client
...
+  # @yieldparam ca_path [String]
+  # @return [void]
+  def with_showcase_tls_groups codepoints
+    # This helper is only safe when tests run one at a time. It repoints the
+    # process-wide SSL_CERT_FILE, which every concurrent REST connection would
+    # pick up, and restores it on exit, which would clobber an overlapping
+    # caller's value. It also binds a fixed port and CA file path, so two
+    # overlapping calls would collide. Minitest runs serially by default;
+    # parallelize_me! switches the class's test_order to :parallel.
+    refute_equal :parallel, self.class.test_order,
```
torreypayne: @amanda-tarafa added the refute assertion here for the parallelization

shared/test/showcase/pqc_test.rb
torreypayne: added on L224!

